PR #90, opened 2011-10-24
Fix in RF matlab interface

cbecker:
I have noticed that there is a hardcoded value that seems to be incorrectly computed. Because of this, the RF model saved to matlab is incorrect and has a missing value. This commit fixes this, which is a bug that would also apply to any other code calling the serialize() function of the RF options.

I have also changed the MATLAB structure type of the RF options to double, since class labels as well as class weights are typically double, plus the fact that serialze() is using std::copy, so the same type (double) is needed for this to be meaningful.

Maybe this means that the RF matlab interface + options serialization needs to be double-checked.

### DIFF
--- a/include/vigra/random.hxx
+++ b/include/vigra/random.hxx
@@ -140,12 +140,18 @@ struct RandomState<TT800>
 
     UInt32 get() const
     {
-        if(current_ == N)
-            generateNumbers<void>();
-            
-        UInt32 y = state_[current_++];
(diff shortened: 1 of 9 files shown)


ukoethe:
We decided to implement concurrency in VIGRA with std::thread instead of OpenMP. Moreover, putting a critical section in the basic random number generation function (i.e. in the innermost loop) may lead to noticeable slowdown. One should create a private random generator for each thread instead.

PR #199, opened 2014-08-07
vigranumpy/src/core/convolution.cxx: Added convolveOneDimension binding ...

jakirkham:
...to pythonConvolveOneDimensionND<float,5> so that it supports convolution over 4D arrays.


stuarteberg:
Is there any reason this PR can't be merged immediately?

PR #201, opened 2014-08-12
WIP: parallel version of multi_pointoperators.hxx

minhb:
Hi,

-I modified the way we update _the number of threads involve in the paralel region_. Currently, all threads update the variable _num_threads_. I think one thread updates this variable should be enough.

-I made the parallel version of multi_pointoperators.hxx. The test file is located in test/openmp_multi_pointoperators folder.  The following functions was parallelized and partially tested: initMultiArray, copyMultiArray and transformMultiArray. The runtime performance gains with very highly dimensional array which has big dimension lenght.

Kind regards,
Minh

### DIFF
--- a/test/openmp/test_omp_multi_pointoperators.cxx
+++ b/test/openmp/test_omp_multi_pointoperators.cxx
@@ -44,13 +44,18 @@
 
 using namespace std;
 using namespace vigra;
+using namespace vigra::functor;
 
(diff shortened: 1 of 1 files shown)


ukoethe:
This PR cannot be used as is because we abandoned OpenMP in favor of std::thread. But it still contains lots of useful code, so I keep it open for now.

PR #202, opened 2014-08-13
Added/extended registration algorithms to/of the VIGRA

BSeppke:
Dear Ulli,

I wanted to get the 2d registration algorithms adapted to vigra-style before going into holidays for 4 weeks (actually, it's my honeymoon). And I did succeed with all algorithms I'd like to add, except the piecewise affine registration, which may require a bit more discussion, since it requires a Delaunay triangulation. So, here is what you get.

1. Affine Registration (image -> image)
---
The algorithms inside the vigra use (local) image gradient information the derive the optical flow for different motion models. The optical flow is then used to determine the affine registration matrix. However, if the displacements are too large, the correspondence between the optical flow and the motion/displacement vanishes or is even misleading. 
For such cases, the new global registration approach is able to still find out 1. if a registration between the two images is possible w.r.t. rotation and translation and 2. gives the resulting affine registration matrix. The algorithm uses the normalized cross correlation and a polar transformation of the FFT of both images to determine the best rotation angle and than uses the fast normalized cross correlation of five subimages of the first (upleft, upright, loleft, loright, center) to be matched at the second image. The matching quality can be derived w.r.t. both correlation coefficients (rotation and translation).

2. Projective Registration (point set -> point set)
--
Works just like the affine registration, but estimates a projective transformation matrix instead of an affine one.

3. Polynomial Registration (point set -> point set)
--
Template based polynomial registration approach of any degree. Although this normally cannot be represented by linear transformations (like matrices), I used matrices to store the resulting coefficients. This ensures a vigra compliant style, since the coefficients can be computed first, and the "polynomial" matrix is then used to perform the registration.

4. Radial Basis Function Registration (point set -> point set)
--
Like the polynomial registration, this is a template based framework where one can use also self designed radial basis functions. I supplied the famous Thin Plate Spline functor and the Distance Power functor (of any degree) as examples. The resulting transformation has to be precomputed as a weight matrix only once (cf. the polynomial registration). It then can be used to warp/register the images.

All of the extensions bring with their own unit-tests (test/registration), which run without any errors (for me). I hope, that you like my extensions and are willing to include them into the vigra. Please note, that I'll be back at office at Spetember, 12th. 

Best wishes
Benjamin


BSeppke:
Hi Ulli,

are you sure, that my pull request is raising the errors inside the TravisCI build? Its been a while since I proposed this pull request and I am still optimistic that this could be a usefull addition to the vigra. However, if I can help you with this pull request, let me know.

BSeppke:
BTW I merged my repository with the current master branch of the vigra and tried to compile - works fine. Here is what the tests of the corresponding registration-TestSuite looks like:
Known transformation: 0.908143 -0.418660 15.788285 
0.418660 0.908143 -15.902489 
0.000000 0.000000 1.000000 

Estimated transformation: 
0.908102 -0.418748 15.799598 
0.418748 0.908102 -15.910708 
0.000000 0.000000 1.000000 
(rot-certainty: 0.957118, trans-certainty: 0.999998)


Known transformation: 
0.911762 -0.410719 233.552028 
0.410719 0.911762 -294.561638 
0.000000 0.000000 1.000000 

Estimated transformation: 
0.912017 -0.410152 233.435538 
0.410152 0.912017 -294.494019 
0.000000 0.000000 1.000000 
(rot-certainty: 0.899532, trans-certainty: 0.990834)


Entering test suite RegistrationTestCollection
Entering test suite EstimateGlobalRotationTranslationTestSuite
All (2) tests passed in test suite EstimateGlobalRotationTranslationTestSuite
Leaving test suite EstimateGlobalRotationTranslationTestSuite
Entering test suite ProjectiveRegistrationTestSuite
All (2) tests passed in test suite ProjectiveRegistrationTestSuite
Leaving test suite ProjectiveRegistrationTestSuite
Entering test suite PolynomialRegistrationTestSuite
All (5) tests passed in test suite PolynomialRegistrationTestSuite
Leaving test suite PolynomialRegistrationTestSuite
Entering test suite RadialBasisRegistrationTestSuite
All (7) tests passed in test suite RadialBasisRegistrationTestSuite
Leaving test suite RadialBasisRegistrationTestSuite
All (16) tests passed in test suite RegistrationTestCollection
Leaving test suite RegistrationTestCollection

So, everything works fine ;)

ukoethe:
Thanks for submitting this PR and remaining patient for so long!

PR #203, opened 2014-08-14
WIP: parallel version of multi_localminmax.hxx

minhb:
Hi,

I updated the parallel version of the node iteration section in the following functions: 
- boost_graph::localMinMaxGraph
- lemon_graph::localMinMaxGraph
- lemon_graph::extendedLocalMinMaxGraph

The test file is _test/openmp/test_omp_multi_localminmax.cxx_

There are two issues I have:
1. I could not find the test for function: _lemon_graph::extendedLocalMinMaxGraph_. So, I tested only two left functions.
2. In the parallel region, I want to get rid of `break` statement. I think, in the parallel region when one thread hits `break` statement, the others should stop executing their loop iterations. This could save us some useless work. 

Although, I would like to make this pull request to discuss about the parallel version of above functions. 

Remark: I would like to make a separate pull request for this issue. Unfortunately, the previous commits still stayed in this pull request. So, please consider changes in only two files: _test/openmp/test_omp_multi_localminmax.cxx_ and _include/vigra/multi_localminmax.hxx_ .

Best regards,
Minh

### DIFF
--- a/include/vigra/multi_localminmax.hxx
+++ b/include/vigra/multi_localminmax.hxx
@@ -50,51 +50,116 @@ namespace vigra {
 
 namespace boost_graph {
 
-  // Attempt without LValue propmaps, using only the free functions
-  // to access ReadablePropertyMap (input) and WritablePropertyMap (label)
+#ifdef OPENMP
+
+//TODO: Get rid of break statement inside the parallel region?
+
+// Attempt without LValue propmaps, using only the free functions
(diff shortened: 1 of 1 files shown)


ukoethe:
This PR cannot be used as is because we abandoned OpenMP in favor of std::thread. But it still contains lots of useful code, so I keep it open for now.

PR #216, opened 2014-10-29
 	Adds offset to cost function (in skeleton::skeletonizeImage)

ilastikdev:
It breaks ties between pixels of equal cost.


ukoethe:
Merged.

Issue #220, opened 2014-11-27
Compilation issues with chunked array on Ubuntu 14.04


hmeine:
On Ubuntu 14.04.1 LTS with GCC 4.6, current master does not compile due to chunked array:

```
[ 60%] Building CXX object vigranumpy/src/core/CMakeFiles/vigranumpy_core.dir/multi_array_chunked.cxx.o                                    
In file included from /home/hmeine/vigra/include/vigra/multi_array_chunked.hxx:140:0,
                 from /home/hmeine/vigra/vigranumpy/src/core/multi_array_chunked.cxx:42:
/home/hmeine/vigra/include/vigra/threading.hxx:92:25: error: ‘std::this_thread::yield’ has not been declared
/home/hmeine/vigra/include/vigra/threading.hxx:94:25: error: ‘std::this_thread::sleep_for’ has not been declared
/home/hmeine/vigra/include/vigra/threading.hxx:95:25: error: ‘std::this_thread::sleep_until’ has not been declared
In file included from /home/hmeine/vigra/vigranumpy/src/core/multi_array_chunked.cxx:42:0:
/home/hmeine/vigra/include/vigra/multi_array_chunked.hxx: In member function ‘long int vigra::ChunkedArray<N, T>::acquireRef(vigra::ChunkedArray<N, T>::Handle*) const’:                                                                                                              
/home/hmeine/vigra/include/vigra/multi_array_chunked.hxx:1635:21: error: ‘yield’ is not a member of ‘vigra::threading::this_thread’
/home/hmeine/vigra/include/vigra/multi_array_chunked.hxx:1635:21: note: suggested alternative:
/usr/include/boost/smart_ptr/detail/yield_k.hpp:97:13: note:   ‘boost::detail::yield’
make[2]: *** [vigranumpy/src/core/CMakeFiles/vigranumpy_core.dir/multi_array_chunked.cxx.o] Fehler 1
```

Apparently, the threading support in the std:: namespace is not complete enough, but the necessary compiler checks for GCC are still missing?  The following looks like someone tried to care about this, but maybe it needs to be extended:
```
#  if defined(__GNUC__) && (!defined(_GLIBCXX_HAS_GTHREADS) || !defined(_GLIBCXX_USE_C99_STDINT_TR1))
#    define VIGRA_NO_STD_THREADING
#  endif
```

PR #224, opened 2015-01-06
Mac libcxx fixes

stuarteberg:
All of vigra (including threading support) can be linked against `libc++` (instead of GNU `libstdc++`) on Mac with these minor changes.  (As before, Mac users who use `libstdc++` will still need to use `-DWITH_BOOST_THREAD=1`.)


stuarteberg:
Sorry, this PR causes problems on linux.  I'll submit a new one when it's ready.

hmeine:
Hey, this reminds me of issue #220, in particular in my comment https://github.com/ukoethe/vigra/issues/220#issuecomment-64899664 I wondered whether we should have a CMake test for the (GCC) compiler config. Would be great if we could kill two birds with one stone..

PR #225, opened 2015-01-07
Mac libcxx fixes (second attempt)

stuarteberg:
I fixed the bad logic that caused linux issues in my original PR.  These changes are now known to work on Mac and not break linux.

All of vigra (including threading support) can be linked against libc++ (instead of GNU libstdc++) on Mac with these minor changes. (As before, Mac users who use libstdc++ will still need to use -DWITH_BOOST_THREAD=1.)


stuarteberg:
BTW: Sorry @hmeine, I made no attempt to "kill two birds with one stone" as [you suggested][1] in the previous PR. 

[1]: https://github.com/ukoethe/vigra/pull/224#issuecomment-68993048

config/check_std_thread.cxx
```diff
@@ -0,0 +1,3 @@
+#include <thread>
+int main() {}
```
hmeine: Would it be possible to extend this test program to call `std::this_thread::yield();`?  Then, we could eventually re-use this for Linux (cf. issue #220).

include/vigra/mathutil.hxx
```diff
@@ -164,6 +164,17 @@ VIGRA_DEFINE_MISSING_ABS(signed long long)
 using std::isinf;
 using std::isnan;
 
+#ifdef _LIBCPP_VERSION
+// On older versions of Mac, the system's libc++ does not provide isnan(double)
+// So we just provide it ourselves.
```
hmeine: Is this addition safe? I know we have isnan() workarounds in other places already, so we should probably agree how to do this in vigra?! For instance, I used `#include <boost/math/special_functions/fpclassify.hpp>` and then `boost::math::isnan` for maximum compatibilty. (I see that `isnan()` may be simple enough not to require boost for that if it is not already a requirement.)

include/vigra/mathutil.hxx
```diff
@@ -164,6 +164,17 @@ VIGRA_DEFINE_MISSING_ABS(signed long long)
...
+template <typename T>
+typename std::enable_if<std::is_integral<T>::value, bool>::type
```
ukoethe: I don't quite understand this. Why do we need `isnan()` for integer types? And if we need it (e.g. for reasons of genericity), why don't we just return `true` then?

include/vigra/mathutil.hxx
```diff
@@ -164,6 +164,17 @@ VIGRA_DEFINE_MISSING_ABS(signed long long)
 using std::isinf;
 using std::isnan;
 
+#ifdef _LIBCPP_VERSION
```
ukoethe: Does this check always imply that no suitable `std::isnan()` is present?

config/check_std_thread.cxx
ukoethe: I think this should be possible.

config/VigraConfigureThreading.cmake
```diff
@@ -2,11 +2,20 @@ MACRO(VIGRA_CONFIGURE_THREADING)
...
+            ${CMAKE_BINARY_DIR} ${PROJECT_SOURCE_DIR}/config/check_std_thread.cxx
+            OUTPUT_VARIABLE THREADING_TEST_OUTPUT )
+        if (NOT THREADING_FOUND)
+            if(${ARGN} MATCHES "REQUIRED")
+                MESSAGE(FATAL_ERROR "Compiler does not support C++ threading. Try to run cmake with '-DWITH_BOOST_THREAD=1'.\n"
+                "For reference, the failed compiler output is shown below."
+                "${THREADING_TEST_OUTPUT}\n\n"
+                "Compiler does not support C++ threading. Try to run cmake with '-DWITH_BOOST_THREAD=1'.\n"
+                "For reference, the failed compiler output is shown above.\n"
+                )
```
ukoethe: Why is the diagnostic given twice?

include/vigra/mathutil.hxx
ukoethe: And more importantly: why does an integer `isnan()` solve the problem of missing `isnan()` for floating types?

config/VigraConfigureThreading.cmake
stuarteberg: I wanted to display the failed compiler output (`THREADING_TEST_OUTPUT`), in case it was useful for the user to figure out *why* his threading implementation was broken, but I was afraid the gobbdelygook from the compiler would obscure the overall diagnostic message.  So, I printed it before and after the output.  But if you think that's confusing, let's just change it.

include/vigra/mathutil.hxx
stuarteberg: In old versions of `libc++`, the `isnan` function is implemented using `std::enable_if<std::is_floating<T>::value, bool>`, but newer versions use `std::enable_if<std::is_arithmetic<T>::value, bool>`, which also supports integers.

I did some investigation here, but came up with no useful check for detecting a useful check for distinguishing between old and new versions of `libc++`.  I am confused about why they designed the library like this, but as far as I can tell, the `_LIBCPP_VERSION` macro always evaluates to `1101`.  (I tried checking out the `libc++` source from their svn repo and compared my version with the current trunk.)  There might be a "proper" solution here, but I don't know what it is.

include/vigra/mathutil.hxx
stuarteberg: Oooops!  The comment here is incorrect.  All versions of `libc++` provide `isnan(double)`.  It is the integer types that cause the problem, and that is what is fixed here.

include/vigra/mathutil.hxx
stuarteberg: >Why do we need isnan() for integer types?

This is the error message I'm seeing:

```
In file included from /miniconda/conda-bld/work/include/vigra/random_forest.hxx:59:
/miniconda/conda-bld/work/include/vigra/random_forest/rf_preprocessing.hxx:131:16: error: no matching
      function for call to 'isnan'
            if(isnan(in[ii]))
               ^~~~~
/miniconda/conda-bld/work/include/vigra/random_forest/rf_preprocessing.hxx:179:29: note: in instantiation
      of function template specialization 'vigra::detail::contains_nan<2, unsigned int,
      vigra::StridedArrayTag>' requested here
        vigra_precondition(!detail::contains_nan(response), "RandomForest(): Response "
                            ^
/miniconda/conda-bld/work/include/vigra/error.hxx:269:81: note: expanded from macro 'vigra_precondition'
#define vigra_precondition(PREDICATE, MESSAGE) vigra::throw_precondition_error((PREDICATE), MESSAGE,...
                                                                                ^
/miniconda/conda-bld/work/include/vigra/random_forest.hxx:985:20: note: in instantiation of function
      template specialization 'vigra::Processor<vigra::ClassificationTag, unsigned int, float,
      vigra::StridedArrayTag, unsigned int, vigra::StridedArrayTag>::Processor<unsigned int>' requested
      here
    Preprocessor_t preprocessor(    features, response,
                   ^
/miniconda/conda-bld/work/vigranumpy/src/core/random_forest.cxx:188:12: note: in instantiation of function
      template specialization 'vigra::RandomForest<unsigned int, vigra::ClassificationTag>::learn<float,
      vigra::StridedArrayTag, unsigned int, vigra::StridedArrayTag, vigra::detail::RF_DEFAULT,
      vigra::DepthAndSizeStopping,
      vigra::rf::visitors::detail::VisitorNode<vigra::rf::visitors::OOB_Error,
      vigra::rf::visitors::StopVisiting>, vigra::RandomNumberGenerator<vigra::detail::RandomState<MT19937>
      > >' requested here
        rf.learn(trainData, trainLabels, visitors::create_visitor(oob_v),
           ^
/miniconda/conda-bld/work/vigranumpy/src/core/random_forest.cxx:407:33: note: in instantiation of function
      template specialization 'vigra::pythonLearnRandomForest<unsigned int, float>' requested here
             registerConverters(&pythonLearnRandomForest<LabelType,float>),
                                ^
/usr/bin/../lib/c++/v1/cmath:423:25: note: candidate template ignored: disabled by 'enable_if'
      [with _A1 = unsigned int]
typename std::enable_if<std::is_floating_point<_A1>::value, bool>::type
                        ^
1 error generated.
make[2]: *** [vigranumpy/src/core/CMakeFiles/vigranumpy_learning.dir/random_forest.cxx.o] Error 1
```

stuarteberg: >And if we need it (e.g. for reasons of genericity), why don't we just return true then?

I considered doing that, but I found [this page][1] which claims that it is theoretically possible for an architecture to specify integer Nan values (though x86 is not one of them).  Maybe it's not worth worrying about such bizarre cases, but using `x != x` should be more correct in such cases.  (The compiler will just optimize it away on most architectures.)

[1]: http://stackoverflow.com/questions/3949457/can-an-integer-be-nan-in-c

stuarteberg: >And more importantly: why does an integer isnan() solve the problem of missing isnan() for floating types?

So sorry for the confusion.  The comment in the code is incorrect.  The old version of `libc++` DOES provide `isnan(double)`.  It fails to provide `isnan(int)`, etc.  Not sure what I was thinking when I wrote that comment....

config/VigraConfigureThreading.cmake
ukoethe: Ok. Ideally, I'd imagine output like this:
```
Checking for threading support:   std::thread 
     or
Checking for threading support:   boost::thread 
     or
Checking for threading support:   not found 
(compiling single threaded, consider running 'cmake -DWITH_BOOST_THREAD=1')
     or
Checking for threading support:   not found
***********************************************
   test program output
***********************************************
C++ threading is required, but your compiler doesn't support it.
For reference, the failed compiler output is shown above.
Try to run cmake with '-DWITH_BOOST_THREAD=1'.
```
It may be a bit of overkill, but you get the idea.

include/vigra/mathutil.hxx
ukoethe:
```
         if(isnan(in[ii]))
```
I'm surprised that this has never failed on any other platform -- is it really conceivable that everyone else implements `isnan(int)`?

It also appears that `std::enable_if<std::is_integral<T>::value, bool>::type` is already supported by Visual Studio 2010. Are there other old compilers that might cause problems here?

stuarteberg: It appears that `isnan` was [added to the std library for C++11][1], so this issue can only appear when using a std library implementation that is new enough to provide C++11 support, but old enough that the C++11 standard was not yet finalized.  I suppose it's possible that old versions of `stdlibc++` could suffer from this issue, but I don't know.

[1]: http://en.cppreference.com/w/cpp/numeric/math/isnan

config/VigraConfigureThreading.cmake
stuarteberg: >Ideally, I'd imagine output like this

Looks reasonable to me.

stuarteberg:
OK, I've added a few commits to this PR that hopefully address all the action items from above, and then some.

- Fixed that erroneous comment
- Added `std::this_thread::yield()` to the `check_std_thread.cxx`
- Cleaned up the error message shown if `std::thread` is broken.
- When checking the threading configuration, the test compile is only performed ONCE, even if the macro is called multiple times from different `CMakeLists.txt` files.
- Fixed linker issue in (previously broken) `test/blockwisealgorithms/CMakeLists.txt`

These changes seem to do the right thing for both values of `-DWITH_BOOST_THREAD`, on both Mac and Linux.  I have not tested on Windows.

hmeine:
Actually, on my I am getting
```
In file included from /Users/hmeine/vigra/vigranumpy/src/core/random_forest.cxx:41:
In file included from /Users/hmeine/vigra/include/vigra/random_forest.hxx:59:
/Users/hmeine/vigra/include/vigra/random_forest/rf_preprocessing.hxx:131:16: error: call to 'isnan' is ambiguous
            if(isnan(in[ii]))
               ^~~~~
```
Apple LLVM version 6.0 (clang-600.0.56) (based on LLVM 3.5svn)
Target: x86_64-apple-darwin13.4.0

stuarteberg:
Closing this PR, and replacing it with #227.

PR #226, opened 2015-01-08
remove outdated documentation link in README

hmeine:
The HCI URL redirects to the github.io one, but the README mentions both
as documenting different versions.

*Note:* The "GridGraph" parts of the online documentation are currently broken / missing (404s).

(Also, beautify the URLs by omitting /index.html.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,12 +36,11 @@ or online at
 Documentation
 -------------
 
-If you downloaded an official release, the documentation can be found in $VIGRA_PATH/doc/vigra/, the start file 
-is $VIGRA_PATH/doc/vigra/index.html. Online documentation for the latest release is at 
-http://hci.iwr.uni-heidelberg.de/vigra/doc/vigra/index.html
+If you downloaded an official release, the documentation can be found in `$VIGRA_PATH/doc/vigra/`, the start file 
+is `$VIGRA_PATH/doc/vigra/index.html`.
 
 When you use the development version from github, you can generate documentation by `make doc`. Up-to-date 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -36,12 +36,11 @@ or online at
...
 
-If you downloaded an official release, the documentation can be found in $VIGRA_PATH/doc/vigra/, the start file 
-is $VIGRA_PATH/doc/vigra/index.html. Online documentation for the latest release is at 
-http://hci.iwr.uni-heidelberg.de/vigra/doc/vigra/index.html
+If you downloaded an official release, the documentation can be found in `$VIGRA_PATH/doc/vigra/`, the start file 
+is `$VIGRA_PATH/doc/vigra/index.html`.
 
 When you use the development version from github, you can generate documentation by `make doc`. Up-to-date 
-online documentation for the 'master' branch is at http://ukoethe.github.io/vigra/doc/vigra/index.html
+online documentation for the 'master' branch is regularly pushed to http://ukoethe.github.io/vigra/doc/vigra/
```
ukoethe: Unfortunately, "is regularly pushed to" is mainly wishful thinking... I recantly read that travis can in principle generate and upload documentation automatically, but it requires to place the encrypted password into the sources. Is this safe?

hmeine: If you place the encrypted password into the sources, anyone can read it, right? So it depends on what the password can do. If one creates a special account only for the docs, danger might be limited to the docs, but that would still mean that anyone could insert malicious code into the web pages. Does not sound like a good idea. How about letting any Linux machine run "git fetch && make doc && ..." in a cron job?

PR #227, opened 2015-01-09
Mac libcxx fixes (third attempt)

stuarteberg:
Hi again,

I've reworked these changes once more.  I believe the build problems on Travis and Mac mentioned in #225 should now be fixed (we'll see what Travis says).

Instead of updating the old PR, I've created a new branch/PR with clean commits for each change.  Please review them.  In particular, my changes to `vigra/config.hxx` should be double-checked.

Other considerations:

- As before, if the user specifies `-DWITH_BOOST_THREAD`, then we never add `-std=c++11` to `CMAKE_CXX_FLAGS`, which means that the entire project is compiled in C++98 mode.  As a result, modules that **require** C++11 (so far, just the multiarray tests) should be skipped.  C++98 mode is now detected in cmake via a new `try_run` step to detect the value of `__cplusplus`
- I don't know how to "test" for a working `isnan(int)` function at compile time, so I used `try_compile` at cmake time, along with a preprocessor definition.


stuarteberg:
BTW: I tested these changes on Mac with both `libstdc++` (an old version), `libc++` (an old version), and both with and without `-DWITH_BOOST_THREAD`.  Also, I tested these changes in Linux (Ubuntu 12.04, gcc-4.7), again both with/without `-DWITH_BOOST_THREAD`.  I did not test on Windows, and I did not test on a newer Mac/libc++ system.

stuarteberg:
**WTF**.  Apparently the version of gcc on Travis reports `__cplusplus` as `1` instead of `199711L`, presumably due to the bug [mentioned here][1].  Apparently this was fixed in gcc-4.7, but not in gcc-4.6.

[1]: http://stackoverflow.com/questions/7530047/gnu-c-macro-cplusplus-standard-conform

stuarteberg:
There, that last commit fixes the issue.

By the way, the vigra [installation instructions][1] say to use `make check` for running the tests, but the vigra [`.travis.yml`][2] file uses `make test`.  Is there a difference?  Does it matter?

[1]: http://ukoethe.github.io/vigra/doc/vigra/Installation.html
[2]: https://github.com/ukoethe/vigra/blob/master/.travis.yml#L15

ukoethe:
`make test` is a predefined target of cmake/ctest. It only works after a successfull `make` call (provided that cmake was configured with `-DAUTOBUILD_TESTS=1`). That is, one first compiles everything and then runs all tests. In contrast, `make check` executes each test immediately after compilation (unless cmake was configured with `-DAUTOEXEC_TESTS=0`). I prefer the latter behavior for test-driven development, whereas the former may be better for automated testing. I tried to implement both behaviors with just `make test`, but couldn't get cmake to do what I wanted.

include/vigra/config.hxx
```diff
@@ -149,7 +149,7 @@
 //                                                       //
 ///////////////////////////////////////////////////////////
 
-#if defined(__GNUC__)
+#if defined(__GNUC__) and !defined(__clang__)
```
ukoethe: Visual Studio wants `&&` instead of `and` here.

ukoethe:
It's a shame how much valuable brain power has to be invested into these weired configuration issues.

include/vigra/config.hxx
```diff
@@ -149,7 +149,7 @@
 //                                                       //
 ///////////////////////////////////////////////////////////
```
ukoethe: Can you add `#define VIGRA_NO_INT_ISNAN` at the appropriate places (i.e. for the platforms where we already know that it is required)? This way, most people using VIGRA as a dependency won't be forced to run your `include(VigraConfigCmath)` in their cmake files.

test/blockwisealgorithms/CMakeLists.txt
```diff
@@ -1,4 +1,14 @@
 VIGRA_CONFIGURE_THREADING()
-VIGRA_ADD_TEST(test_blockwiselabeling test_labeling.cxx)
-VIGRA_ADD_TEST(test_blockwisewatersheds test_watersheds.cxx)
-VIGRA_ADD_TEST(test_blockwiseconvolution test_convolution.cxx)
+if(NOT THREADING_FOUND)
```
ukoethe: It would be easy (if ugly) to add support for single threaded builds to the chunked array classes (in fact, I already had it in an earlier version). Is this extension worthwhile?

test/multiarray/CMakeLists.txt
```diff
@@ -1,42 +1,59 @@
...
+endif()
 
-if(NOT THREADING_FOUND)
-    MESSAGE(STATUS "** WARNING: Your compiler does not support C++ threading.")
-    MESSAGE(STATUS "**          test_multiarray_chunked will not be executed on this platform.")
-    if(NOT WITH_BOOST_THREAD)
-        MESSAGE(STATUS "**          Try to run cmake with '-DWITH_BOOST_THREAD=1' to use boost threading.")
-    endif()
-    
+# Multiarray source files use 'auto' and other features from c++11.
```
ukoethe: Really? I don't see any `auto` there, and it would be a bug. The plain multiarray test should alwas be executed, regardless of the threading configuration.

test/multiarray/CMakeLists.txt
```diff
@@ -1,42 +1,59 @@
...
+string(COMPARE LESS ${VIGRA_CPP_VERSION} "201103" NO_CXX11)
+if(NO_CXX11)
+    MESSAGE(STATUS "** WARNING: You are compiling in C++98 mode.")
```
ukoethe: This is probably too simplistic on Windows: `__cplusplus == 199711` for Visual Studio 2010 and 2012, although both provide partial support for C++11 (and I was careful to use only the supported subset so far).

config/check_std_isnan.cxx
```diff
@@ -0,0 +1,8 @@
+#include <cmath>
+int main()
+{
+    // Some old versions of libc++ do not provide isnan() for integers,
+    //  in which case this will fail to compile.
+    std::isnan(int(0));
```
ukoethe: This is strange: why doesn't the compiler perform an implicit conversion to `double` here? In fact, cppreference (http://en.cppreference.com/w/cpp/numeric/math/isnan) explicitly says that this would happen. Maybe the actual problem is something else?

include/vigra/mathutil.hxx
```diff
@@ -164,6 +164,17 @@ VIGRA_DEFINE_MISSING_ABS(signed long long)
 using std::isinf;
 using std::isnan;
 
+#ifdef VIGRA_NO_INT_ISNAN
```
ukoethe: Maybe this condition can be dropped altogether? The template will only match if there is no explicit implementation of `isnan` for a given type, and then it will do the right thing.

include/vigra/config.hxx
```diff
@@ -162,18 +162,57 @@
...
+        #define VIGRA_SHARED_PTR_IN_TR1
+    #endif
+#endif  // __GNUC__
+
+///////////////////////////////////////////////////////////
+//                                                       //
+//                         clang                         //
+//                                                       //
+///////////////////////////////////////////////////////////
+#if defined(__clang__)
```
ukoethe: The following looks ok to me.

include/vigra/config.hxx
stuarteberg: I don't know how to detect the platforms that need it without that test file.  The important factor is the user's particular copy of `libc++`, but not all versions of `libc++` are affected.  Do you know of a way to ask CMAKE what version of the C++ std library is active?

include/vigra/mathutil.hxx
stuarteberg: @hmeine [reported][1] that this extra definition of `isnan` causes errors if the std library already provides its own.  That's why I added this condition.

[1]: https://github.com/ukoethe/vigra/pull/225#issuecomment-69341499

test/blockwisealgorithms/CMakeLists.txt
stuarteberg: I'm not sure, but probably not.  On most platforms, threading support is available.  My guess is that people with old compilers are probably more interested in bug fixes for the features they already use than in new features like the chunked array.

test/multiarray/CMakeLists.txt
stuarteberg: OK, The comment is not quite right.  The uses of `auto` I was referring to are in `test/multiarray/test.cxx`:

```
$ grep -r auto test/multiarray/
test/multiarray/CMakeLists.txt:# Multiarray source files use 'auto' and other features from c++11.
test/multiarray/test.cxx:        auto i3_f = createHierarchicalIterator(a3),
test/multiarray/test.cxx:        auto i3_c = zip(i3_f, i3_f);
test/multiarray/test.cxx:            auto i2_f = i3_f.begin ();
test/multiarray/test.cxx:            auto i2_c = i3_c.begin ();
test/multiarray/test.cxx:            auto i2_l = i3_f.end ();
test/multiarray/test.cxx:                auto i1_f = i2_f.begin ();
test/multiarray/test.cxx:                auto i1_c = i2_c.begin ();
test/multiarray/test.cxx:                auto i1_l = i2_f.end ();
```

config/check_std_isnan.cxx
stuarteberg: Implicit conversion doesn't happen because `enable_if`is used.  Here's the `isnan` implementation in my outdated copy of `libc++`:

```
template <class _A1>
inline _LIBCPP_INLINE_VISIBILITY
typename std::enable_if<std::is_floating_point<_A1>::value, bool>::type
isnan(_A1 __x) _NOEXCEPT
{
    return __libcpp_isnan(__x);
}
```

include/vigra/config.hxx
stuarteberg: Thanks, now fixed.

test/multiarray/CMakeLists.txt
stuarteberg: OK, added a special check for windows.

include/vigra/mathutil.hxx
ukoethe: Understood, but unfortunately he did not report the candidate matches to investigate this further. @hmeine, can you add this information?

I still don't like the fact that we need cmake to find out if `VIGRA_NO_INT_ISNAN` needs to be defined, but I cannot suggest a better way because I have no access to a platform where it is required.

ukoethe: I'm leaning to the conclusion that it is much better to change `rf_preprocessing.hxx` to call
```
      if(isnan(NumericTraits<T>::toRealPromote(in[ii])))
```
to solve the problem where it occurs.

ukoethe: I just realize that the functions `contains_nan` and `contains_inf` need polishing anyway, as they use the potentially slow `in[ii]` (with `ii` being interger) to access the data.

hmeine: That's exactly the conclusion that I came to – instead of fighting hard to provide isnan(int) when it is not present, just call an overload variant that is definitely there.  `toRealPromote` sounds like a good plan, no?  That would simplify things a lot, and I don't think that we support any existing platform where isnan(int) is not always true.  (I can well imagine stuff like fixed-point decimals supporting NaN, but that's again a different topic, and could be supported via an appropriate isnan()-overload and toRealPromote.)

ukoethe: Ok, I changed it this way and removed the `isnan()` configuration code.

stuarteberg:
Thanks for your attention on this, @ukoethe and @hmeine.  Much appreciated!

PR #239, opened 2015-05-01
Use Travis CI's version of the Python library

jakirkham:
Travis CI installs Python under the `/opt` directory instead. This looks for `libpython` in that directory. Additionally, Travis CI sets the environment variable `$TRAVIS_PYTHON_VERSION`. We can leverage this value to get at the Python library. This will be useful when using build matrices to test compatibility on other versions of Python (e.g. 3.x).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,6 @@ before_script:
    
 script:
   - mkdir build && cd build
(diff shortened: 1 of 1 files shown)


jakirkham:
@ukoethe, could you please take a look at this when you have a chance? It is a very simple change and does not even change VIGRA. It only ensures Travis CI links against the right Python library. Currently, the wrong Python library is being linked in Travis CI. The correction location is under `/opt/python` as can be seen here ( https://github.com/travis-ci/travis-cookbooks/blob/master/ci_environment/python/recipes/pyenv.rb#L27 ). This is demonstrated by the test result, as well.

hmeine:
Thanks for the reminder; the pull request looks totally sane.

jakirkham:
Thanks @hmeine.

PR #240, opened 2015-05-08
cmake: Added VIGRA_IMPEX_EXTRA_C_FLAGS variable

stuarteberg:
To get vigra to build on my special CentOS-5.11 VM (with special `gcc` binary), I needed to add some extra flags for the C files in `impex`.  This PR adds a variable to allow users to add compiler flags specifically for those C files.  (It shouldn't be needed on most systems.)

In case you're curious, the extra flags I'm using are:

    -DVIGRA_IMPEX_EXTRA_C_FLAGS="-std=gnu90 -Wno-pedantic -Wno-long-long"

### DIFF
--- a/src/impex/CMakeLists.txt
+++ b/src/impex/CMakeLists.txt
@@ -61,6 +61,12 @@ ADD_LIBRARY(vigraimpex ${LIBTYPE}
     viff.cxx
     void_vector.cxx)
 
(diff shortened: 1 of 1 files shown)


stuarteberg:
Nevermind this PR.  I've made some changes to my compiler setup, so this hacky option isn't needed any more.

PR #242, opened 2015-05-27
Fix cmake issue when OpenMP is present

stuarteberg:
On mac, somehow the test program executed by `VigraDetectCppVersion.cmake` can fail to execute if it was compiled with `-fopenmp` but then can't locate `libgomp.dylib` at runtime.  This simple PR just re-orders cmake statements so that the CPP version is detected before `-fopenmp` is added to `CMAKE_CXX_FLAGS`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -95,17 +95,6 @@ VIGRA_FIND_PACKAGE(FFTW3 NAMES libfftw3-3 libfftw-3.3)
 VIGRA_FIND_PACKAGE(FFTW3F NAMES libfftw3f-3 libfftwf-3.3)
 
 
-#FIND_PACKAGE( OpenMP REQUIRED)
-FIND_PACKAGE( OpenMP )
-if(OPENMP_FOUND)
-message("OPENMP FOUND")
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${OpenMP_C_FLAGS}")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
-set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${OpenMP_EXE_LINKER_FLAGS}")
-endif()
-
-
-
 IF(WITH_OPENEXR)
     VIGRA_FIND_PACKAGE(OpenEXR)
 ENDIF()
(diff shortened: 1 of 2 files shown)


stuarteberg:
The Travis failure shown above is unrelated to this PR.  The failed test was `test_sampler`.

```
98% tests passed, 1 tests failed out of 48

Total Test time (real) = 250.97 sec

The following tests FAILED:
	 37 - test_sampler (Failed)
Errors while running CTest
make: *** [test] Error 8

The command "make test" exited with 2.

Done. Your build exited with 1.
```

hmeine:
Is that test output reproduced on your machine or copied from Travis (where I saw it, too)?

I tried your branch on my OS X 10.9.5 machine, and found that the detected C++ version changes from 201103 to 199711 (apart from being output much earlier, as expected), leading to the warning that the Multiarray tests will be skipped.  That does not seem to be intended, is it?

I see no change in the OpenMP-related output (about 18 different flags seem to be tested, all "Failed").  Should I?

Compiler is /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ from XCode 6.2.

ukoethe:
The proposed change is problematic. VigraDetectThreading contains code that may change the C++ version, for example:

    SET(CMAKE_CXX_FLAGS "-std=c++11 ${CMAKE_CXX_FLAGS}")

So, there is a good reason to check for the version after VigraDetectThreading has completed. One could argue that silently switching to a different version of the C++ standard is wrong to begin with, but moving VigraDetectCppVersion before VigraDetectThreading makes the confusion worse.

Regarding the travis failure: This might be caused by an unfortunate choice of random numbers in the test. Lets see if it just disappears on the next run.

stuarteberg:
>The proposed change is problematic. VigraDetectThreading contains code that may change the C++ version

Yes, I see that now.  Fixed in my local branch, testing now.  Will push soon.

stuarteberg:
OK, thanks for reviewing the change.  Sorry I didn't notice the issue.  I just pushed a revised changeset, so it should be ready to go after Travis finishes.

BTW, @hmeine:

>I see no change in the OpenMP-related output (about 18 different flags seem to be tested, all "Failed"). Should I?

The OpenMP-related output should not change.  The OpenMP dependency was added in a previous commit (0822c132b4850162e96139b9ba81ddceb1ab8f91).  This PR addresses a complication that can arise as a result.  If OpenMP is not detected on your Mac, I assume you must be using clang, which doesn't support OpenMP yet.  (To use OpenMP on my Mac, I compile with gcc-4.8.)

stuarteberg:
OK, Travis is happy, and the C++ version is once again reported as `201103`.

hmeine:
Looks fine, and works.  Tested with clang and gcc (4.8 and 4.7) on OS X.  Indeed, OpenMP is detected with both GCC versions.  The trying could be less verbose, though.

Issue #243, opened 2015-05-27
CMake policy violated (CMP0026)


hmeine:
CMake 3.0.2 produces a lot of warnings, referring to this issue:

> CMP0026
> -------
> 
> Disallow use of the LOCATION target property.
> 
> CMake 2.8.12 and lower allowed reading the LOCATION target
> property (and configuration-specific variants) to
> determine the eventual location of build targets.  This relies on the
> assumption that all necessary information is available at
> configure-time to determine the final location and filename of the
> target.  However, this property is not fully determined until later at
> generate-time.  At generate time, the $\<TARGET_FILE\> generator
> expression can be used to determine the eventual LOCATION of a target
> output.
> 
> Code which reads the LOCATION target property can be ported to use the
> $\<TARGET_FILE\> generator expression together with the file(GENERATE)
> subcommand to generate a file containing the target location.
> 
> The OLD behavior for this policy is to allow reading the LOCATION
> properties from build-targets.  The NEW behavior for this policy is to
> not to allow reading the LOCATION properties from build-targets.
> 
> This policy was introduced in CMake version 3.0.  CMake version
> 3.0.2 warns when the policy is not set and uses OLD behavior.  Use
> the cmake_policy command to set it to OLD or NEW explicitly.

ukoethe:
Yes, I know. But the explanations in the cmake documentation (partly cited above) are quite bad, so that I'm unable to find out how to fix this. Suggestions?

hmeine:
I am not familiar with CMake at all; I would rather hope for someone else to chime in.

stuarteberg:
I just opened a PR to hide this warning.  On mac, one more cmake policy warning remains, but I've opened a separate issue for that (#245).

ukoethe:
Fixed in 78ed77635754837410c4743c534b2cf4d0f2d3a6

ArchangeGabriel:
You’re running into troubles soon though now:
```
CMake Deprecation Warning at config/VigraAddTest.cmake:66 (cmake_policy):
  The OLD behavior for policy CMP0026 will be removed from a future version
  of CMake.

  The cmake-policies(7) manual explains that the OLD behaviors of all
  policies are deprecated and that a policy should be set to OLD only under
  specific short-term circumstances.  Projects should be ported to the NEW
  behavior and not rely on setting a policy to OLD.
```
They were 56 such messages during build. So a proper fixing will soon be required.

PR #244, opened 2015-06-01
cmake: VigraAddTest: Use OLD policy for CMP0026

stuarteberg:
This PR simply adds an explicit `cmake_policy` statement to `VigraAddTest.cmake`, so that the warning in #243 is disabled.  An alternative fix would be to rewrite a few lines of `VigraAddTest.cmake` to use the `NEW` cmake behavior (and then specify `cmake_policy(SET CMP0026 NEW)`), but (1) I'm not familiar with how to do that, and (2) I think it would require users to upgrade to CMake 3.0.

### DIFF
--- a/config/VigraAddTest.cmake
+++ b/config/VigraAddTest.cmake
@@ -13,6 +13,10 @@
 # The function VIGRA_COPY_TEST_DATA copies the given files from the current source directory
 # to the corresponding binary directory.
 #
+if (NOT (CMAKE_VERSION VERSION_LESS 3.0))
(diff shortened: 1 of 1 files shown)


hmeine:
Maybe `cmake_policy(SET CMP0026 NEW)` requires CMake 3.0, but would fixing the reported problem? I read the explanation like "you should have never done it this way, we just did not give a warning", which would imply that there has always been a better way.

ukoethe:
This problem was fixed for real in 78ed77635754837410c4743c534b2cf4d0f2d3a6.

PR #246, opened 2015-06-10
python: colors.cxx: Added int8 type to a few function overload lists.

stuarteberg:
The functions in `vigranumpy/src/core/colors.cxx` already supported unsigned types `uint8`, `uint16`, and `uint32`, also signed `int16` and `int32`, but signed `int8` was missing for some reason.

(This change is necessary for ilastik to display `int8` grayscale images.)

### DIFF
--- a/vigranumpy/src/core/colors.cxx
+++ b/vigranumpy/src/core/colors.cxx
@@ -496,7 +496,7 @@ void defineColors()
 
     docstring_options doc_options(true, true, false);
 
-    multidef("applyColortable", pyApplyColortable<vigra::UInt8, vigra::Int16, vigra::UInt16, vigra::Int32, vigra::UInt32>(),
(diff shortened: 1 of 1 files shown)


hmeine:
That's probably because `int8` is very unusual.  If you need it in ilastik, that's justification enough, but do you really need it?  Usually, 8bit data is unsigned, people using signed integers tend to always use 16+ bits as well.

stuarteberg:
>Usually, 8bit data is unsigned, people using signed integers tend to always use 16+ bits as well.

I discovered the issue when loading a [test image](http://www.openmicroscopy.org/Schemas/Samples/2015-01/bioformats-artificial/multi-channel-4D-series.ome.tif.zip) from the [OME-TIFF website](http://www.openmicroscopy.org/site/support/ome-model/ome-tiff/data.html).  That image uses `int8` as the pixel type.  Perhaps their example data is poorly chosen.  Nonetheless, it is nice if ilastik can load the examples.

PR #249, opened 2015-07-02
Reset MultiArrayView

sstefan:
In some situations, I'd like to have the possibility to reset a
MultiArrayView object to the same state as after default construction
(i.e., an invalid MultiArrayView that can be assigned to).  

This is convenient if one uses MultiArrayViews objects (e.g. as
members of another object), and these need to change during the object
lifetime to point to different image data, for instance. Currently, I
don't see another way to "reset" an existing MultiArrayView object,
because assignment is defined such that it will either copy the data
(if shape matches) or fail, unless the object is in the "not
initialized yet" state as after default construction. Or am I
overlooking something?

Alternatives to the proposed implementation would perhaps be 
- to allow assignment of a default-constructed MultiArrayView unconditionally to perform the "reset";
- to make method reset(const MultiArrayView &rhs) perform the assignment right away.
- There's already method copy() to copy the data pointed to by the MultiArrayViews. One could therefore perhaps also opt to make operator= only perform a shallow copy of the pointer and shape/strides vectors (but that would of course have the severe drawback to alter existing semantics instead of just extending the interface of the class).

What do you think?

### DIFF
--- a/include/vigra/multi_array.hxx
+++ b/include/vigra/multi_array.hxx
@@ -832,6 +832,16 @@ public:
         return MultiArrayView<N, T, StridedArrayTag>(m_shape, m_stride, m_ptr);
     }
 
+	/** Reset this <tt>MultiArrayView</tt> to an invalid state (as after default construction).
+		Can e.g. be used prior to assignment to make a view object point to new data.
+         */
+    void reset() {
+	m_shape = diff_zero_t(0);
+	m_stride = diff_zero_t(0);
+	m_ptr = 0;
+    }
+
+
         /** Assignment. There are 3 cases:
 
             <ul>
--- a/test/multiarray/test.cxx
+++ b/test/multiarray/test.cxx
@@ -449,7 +449,10 @@ public:
             std::string message(c.what());
             should(0 == expected.compare(message.substr(0,expected.size())));
         }
+	array.reset();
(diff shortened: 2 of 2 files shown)


hmeine:
I liked this as well, but I wondered whether `reset` is unambiguous enough. Maybe it should be called `resetToInvalid` or `resetView`, in order to make more clear that this does not reset the image data or so?

ukoethe:
I was contemplating `release()` as in the smart pointer classes?

hmeine:
Good idea, too.  Then again, there's one thing that MultiArrayView does *not* do: hold data.

BTW: What happens if you call `reset()` on a MultiArray?  Is this a case of violated is-a semantics?

ukoethe:
> BTW: What happens if you call |reset()| on a MultiArray? Is this a case
> of violated is-a semantics?

Good point, should be forbodden.

sstefan:
Would commit e5e54ea4f573a4eaf7347e6bcbe04e05e9f62117 be enough to prevent accidentally calling it on MultiArrays? There's of course still array.view().resetView() which will break things...

hmeine:
Looks fine to me.  Quite a minimal (i.e. beautiful!) approach, but I don't see why that should not work. (With which compiler(s) did you test that?)

PR #251, opened 2015-07-15
Fix shuffle algorithm in RF VariableImportanceVisitor

stuarteberg:
Unless I'm mistaken, the `VariableImportanceVisitor` has a tiny bug.  Computing the feature importances involves shuffling each column of the feature matrix, but the algorithm it uses to perform the shuffle isn't quite ideal.

It looks like the author attempted to implement the [Fisher-Yates shuffle](https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle), but got one detail wrong.  I can't say exactly how this will affect the output of the shuffle, but it probably results in some permutations being more likely than others (i.e. not a true shuffle).

For example, in the broken version, I *think* elements near the beginning of the column are more likely to stay near the beginning of the column after the shuffle is complete, instead of being equally likely to end up in any position throughout the array.


stuarteberg:
BTW, Instead of this fix, an even better fix might be to use `std::random_shuffle` (assuming the appropriate iterators can be obtained for each column of the feature matrix -- I would have to spend some time with the vigra docs)

ukoethe:
Thanks for spotting this error. I don't think that random_shuffle() is needed to improve the design - there is nothing wrong with a simple loop.

PR #254, opened 2015-07-26
Elminate FutureWarning in vigranumpy colors module

stuarteberg:
This tiny change eliminates the following warning from numpy:

    FutureWarning: comparison to `None` will result in an elementwise object comparison in the future.

### DIFF
--- a/vigranumpy/src/core/colors.cxx
+++ b/vigranumpy/src/core/colors.cxx
@@ -382,7 +382,7 @@ void pythonGray2QImage_ARGB32Premultiplied(
     TmpType pixelF;
     
     TmpType normalizeLow, normalizeHigh; 
-    if(normalize != boost::python::object())
+    if(normalize.pyObject() != Py_None)
     {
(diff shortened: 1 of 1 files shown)


vigranumpy/src/core/colors.cxx
```diff
@@ -382,7 +382,7 @@ void pythonGray2QImage_ARGB32Premultiplied(
     TmpType pixelF;
     
     TmpType normalizeLow, normalizeHigh; 
-    if(normalize != boost::python::object())
+    if(normalize.pyObject() != boost::python::object().ptr())
```
hmeine: Thanks a lot!
I think I would have compared with ``Py_None``, wouldn't you prefer that?
(Not only that it's shorter, but it also does not create a temporary.) Anyhow, would be safe to merge.

hmeine:
Nice.  Would be safe to merge, but see my inline comment.

stuarteberg:
>I think I would have compared with Py_None, wouldn't you prefer that?

Agreed.  Now fixed.

PR #257, opened 2015-08-31
Added check for OSX

denics:
In OSX check must be included from AsserMacros OSX library


hmeine:
What for?  Can you describe a failing situation without this change?
(Several vigra users are on OS X, as I am myself, and I have never needed this.)

denics:
I am trying to compile last hugin source and it simply fail compilation when calling check()

PR #267, opened 2015-10-14
Kernel2D: added initExplicitly(BasicImage)

Empage:
Is there a way of initializing a custom kernel at runtime? 

The `initExplicitly` function cannot be used because it requires a comma-separated initializer list, which is why I came up with an overload to take a BasicImage which is then used as the kernel. Currently only odd kernel sizes are allowed to avoid the aligning ambiguities of even sized kernels. 

I could add support for MultiArray, even sized kernels, etc. But first, I wanted to show you the idea to see if you think, it's useful (or whether I missed an already available solution).

### DIFF
--- a/include/vigra/stdconvolution.hxx
+++ b/include/vigra/stdconvolution.hxx
@@ -1154,6 +1154,33 @@ public:
     {
         return initExplicitly(Shape2(upperleft), Shape2(lowerright));
     }
+	
+        /** Init the kernel by providing a BasicImage with the kernel values.
+
+            The norm is set to the sum of the image values. 
+
+            <b> Preconditions:</b>
+
+            odd image width and height;
+        */
+    Kernel2D & initExplicitly(BasicImage<value_type> const & image)
+    {
+        vigra_precondition(image.width() % 2 != 0 && image.height() % 2 != 0,
+                           "Kernel2D::initExplicitly(): kernel sizes must be odd.");
+
+        left_  = Point2D((image.width() - 1) / -2, (image.height() - 1) / -2);
+        right_ = Point2D((image.width() - 1) /  2, (image.height() - 1) /  2);
+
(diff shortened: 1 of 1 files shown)


ukoethe:
At present, the desired functionality can be achieved by
```c++
Point2D left(...), right(...);
kernel.initExplicitly(left, right);
copyImage(srcImageRange(weights), destIter(kernel.center()+left));
```
but a convenience function is, of course, much better. 

Yet another possibility is to overload the constructor and/or assignment operator so that one can write
```c++
Point2D left(...), right(...);
kernel.initExplicitly(left, right) = weights;  // checks that 'weights' has correct shape
```
or just
```c++
Kernel2D<double> kernel(weights);  // sets 'left' and 'right' symmetrically
```
If you are planning to apply big kernels or many kernels, you may also want to compute the convolution in the Fourier domain using `convolveFFT()` or `convolveFFTMany()` from `multi_fft.hxx`.

include/vigra/stdconvolution.hxx
```diff
@@ -1154,6 +1154,33 @@ public:
...
+        */
+    Kernel2D & initExplicitly(BasicImage<value_type> const & image)
+    {
+        vigra_precondition(image.width() % 2 != 0 && image.height() % 2 != 0,
+                           "Kernel2D::initExplicitly(): kernel sizes must be odd.");
+
+        left_  = Point2D((image.width() - 1) / -2, (image.height() - 1) / -2);
+        right_ = Point2D((image.width() - 1) /  2, (image.height() - 1) /  2);
+
+		norm_ = 0;
```
ukoethe: Never use tabs for indentation. VIGRA always indents by 4 spaces.

ukoethe:
Merged into branch 'version-1-11-rc'

PR #272, opened 2015-11-10
Cleanup

DerThorsten:
Changes:
 - fixed bug in cluster operators ('weighted' wardness was wrong)
 - exported shock-filter2d to python
 - made blockwise watershed parallel 
 - added stand alone accumulator chain
 - shortest path segmentation uses the shortest path class instead of reimplementing it by itself
 - added code to find all 3-cycles in a graph
 - made multi histograms better and code cleaner
 - added "gaussian rank" for multiarrays
 - improved python bindings of multi blocking 
 - added some classes for grah algorithms  (as array map) (  These classes will most probably  not be used by end users)
 - replaced OpenMP with threadpool
 - implemented counting iterator (very useful with for_each)
 - improved the threadpool, serial calls are now possible (for debugging)


include/vigra/threadpool.hxx
```diff
@@ -47,21 +47,68 @@
...
+
 
 class ThreadPool {
 public:
 
     /**
+     * Create a thread pool from ParallelOptions
+     */
+    ThreadPool(const ParallelOptions & options)
+    : ThreadPool(options.getNumThreads()){
```
ukoethe: Delegating constructors unsupported on Travis and appveyor

DerThorsten: fixed

DerThorsten:
Travis failure reason..
"The command "sudo apt-get install --assume-yes libhdf5-serial-dev libboost-python-dev libjpeg-dev libtiff4-dev libpng12-dev libfftw3-dev cmake" failed and exited with 100 during ."

DerThorsten:
E: Unable to locate package libhdf5-serial-dev
@ukoethe : it seems that travis is (currently) broken?!?


E: Unable to locate package libboost-python-dev

E: Unable to locate package libfftw3-dev

The command "sudo apt-get install --assume-yes libhdf5-serial-dev libboost-python-dev libjpeg-dev libtiff4-dev libpng12-dev libfftw3-dev cmake" failed and exited with 100 during .

.travis.yml
```diff
@@ -13,3 +13,4 @@ script:
   - cmake -DCMAKE_BUILD_TYPE=Release -DDEPENDENCY_SEARCH_PREFIX=$VIRTUAL_ENV -DAUTOBUILD_TESTS=1 -DAUTOEXEC_TESTS=0 -DVIGRANUMPY_LIBRARIES="/opt/python/${TRAVIS_PYTHON_VERSION}*/lib/libpython${TRAVIS_PYTHON_VERSION}.so;/usr/lib/libboost_python.so" ..
   - make
   - make test
+  - ./test/blockwisealgorithms/test_blockwiselabeling
```
ukoethe: Why is this explicitly called? Should be executed as part of `make test`.

DerThorsten: This was to debug on travis.
Normal test suite does not show enough errors /couts.
Should I remove and commit? Or how should we handle this =)

ukoethe: I'll clean it up myself. Just wanted to make sure.

DerThorsten: could we modify travis, such that we get more verbose output in case there was an error?

ukoethe: It might be possible to keep track of the failed tests and re-run them with more verbose output. But this doesn't sound like it is easy to do.

include/vigra/blockwise_labeling.hxx
```diff
@@ -88,7 +92,8 @@ blockwiseLabeling(DataBlocksIterator data_blocks_begin, DataBlocksIterator data_
                   LabelBlocksIterator label_blocks_begin, LabelBlocksIterator label_blocks_end,
                   NeighborhoodType neighborhood, Equal equal,
                   const Value* background_value,
-                  Mapping& mapping)
+                  Mapping& mapping,
+                  const int nThreads = ParallelOptions::Auto)
```
ukoethe: Is there a way to set the number of threads in the highlevel API of `labelMultiArrayBlockwise()`?

DerThorsten: One could just added, so far I added it just to the HL function of watershed.
But should be easy as pie

.travis.yml
DerThorsten: We could see if there was any error, and in that case run ctests with most verbose flag

include/vigra/blockwise_labeling.hxx
```diff
@@ -108,21 +113,52 @@ blockwiseLabeling(DataBlocksIterator data_blocks_begin, DataBlocksIterator data_
...
+            [&](const int threadId, const uint64_t i){
+                int resVal;
+                if(background_value){
+                    resVal = 1 + labelMultiArrayWithBackground(data_blocks_it[i], label_blocks_it[i],
+                                                                        neighborhood, *background_value, equal);
+                }
+                else{
+                    resVal = labelMultiArray(data_blocks_it[i], label_blocks_it[i],
+                                                      neighborhood, equal);
+                }
```
ukoethe: One of the assignments to `resVal` is probably wrong. Both `labelMultiArrayWithBackground()` and `labelMultiArray()` return the largest valid label. So, `1 +` should be used either in both cases or in neither case.

DerThorsten: In the case it is wrong, it was wrong before I added changes!
Just made the thing parallel...

ukoethe:
@DerThorsten: please follow-up at PR https://github.com/ukoethe/vigra/pull/278

PR #273, opened 2015-11-17
Added threadpool.

dagophil:
Added Thorstens thread pool implementation.


include/vigra/threadpool.hxx
```diff
@@ -0,0 +1,328 @@
...
+    void waitFinished()
+    {
+        std::unique_lock<std::mutex> lock(queue_mutex);
+        conditionF.wait(lock, [this](){ return tasks.empty() && (busy == 0); });
+    }
+
+    size_t nThreads()const{
+        return workers.size();
+    }
+private:
```
ukoethe: Shouldn't the class have a `cancel()` method that drops all tasks not yet running and waits for the others? If yes, the destructor should be changed to call it as well.

include/vigra/threadpool.hxx
```diff
@@ -0,0 +1,328 @@
...
+
+    // the task queue
+    std::queue< std::function<void(int)> > tasks;
+    
+    // synchronization
+    std::mutex queue_mutex;
+    std::condition_variable condition;
+    std::condition_variable conditionF;
+
+    bool stop;
```
ukoethe: If `stop` were atomic, one wouldn't need the lock to set or check the flag. This looks preferable to me.

include/vigra/threadpool.hxx
```diff
@@ -0,0 +1,328 @@
...
+void parallel_foreach_impl(
+    ThreadPool & pool,
+    const uint64_t nItems,                   
+    ITER iter, 
+    ITER end, 
+    F && f,
+    std::random_access_iterator_tag
+){
+    // typedef typename std::iterator_traits<ITER>::reference ReferenceType;
+    uint64_t workload = std::distance(iter, end);
```
ukoethe: Since `nItems` must be passed anyway, I don't see the point of recomputing it.

include/vigra/threadpool.hxx
```diff
@@ -0,0 +1,328 @@
...
+            }
+        );
+        if(workload==0)
+            break;
+        std::advance(iter, lc);
+    }
+    pool.waitFinished();
+}
+
+
```
ukoethe: We might want to add a `parallel_foreach_impl` overload for `std::input_iterator_tag`. It would just assign tasks to each thread in turn. Normally, this would be undesirable due to locality, but for input iterators, it actually makes sense -- why wait for inputs that haven't arrived before utilizing the next thread?

include/vigra/threadpool.hxx
```diff
@@ -0,0 +1,328 @@
...
+template<class ITER, class F>
+void parallel_foreach(
+    ThreadPool & pool,
+    const uint64_t nItems,
+    ITER begin, 
+    ITER end, 
+    F && f
+){
+
+    if(pool.nThreads()!=1){
```
ukoethe: Should the function check the precondition `pool.nThreads() > 0`?

include/vigra/threadpool.hxx
```diff
@@ -0,0 +1,328 @@
...
+void parallel_foreach(
+    const int64_t nThreads,                  
+    const uint64_t nItems,
+    ITER begin, 
+    ITER end, 
+    F && f
+){
+    ThreadPool pool(nThreads==-1 ?  std::thread::hardware_concurrency() : nThreads);
+
+    if(pool.nThreads()!=1){
```
ukoethe: Call this other overload instead of repeating the code. Both overloads should probably be inline.

ukoethe:
@DerThorsten : please review my comments on @dagophil's PR

include/vigra/threadpool.hxx
DerThorsten: no

include/vigra/threadpool.hxx
DerThorsten: ...good point (but in this very casse std::distance is for free since it is a rand access iterator) but still....should be removed

include/vigra/threadpool.hxx
DerThorsten: I do not get this...
What is the problem with the current impl.?

include/vigra/threadpool.hxx
DerThorsten: why not

include/vigra/threadpool.hxx
dagophil: It does not matter whether `stop` is atomic or not. The lock is needed for the condition variable that wakes up waiting threads.

PR #277, opened 2015-12-14
WIP: Port vigranumpy to python 3.5

FynnBe:
Migration to Python 3.x in progress
the quick and dirty solution is about the Python include path that otherwise needs to be set manually for now.


jakirkham:
It seems this build failure on Travis CI is unrelated to the content of this PR. I think it is probably caused by Travis' migration to GCE for hosting their VMs. If given another kick, I expect this would probably pass given Appveyor passed and Travis was passing on the previous commit.

ukoethe:
> I expect this would probably pass given Appveyor passed and Travis was passing on the previous commit.

I don't think so. At the moment, AppVeyor unfortunately doesn't test vigranumpy at all  (due to the lack of a package manager for the dependencies -- we will evatually fix this using conda), and the PR does not even work on our local machines. We nonetheless published the code as a PR already so that people can point out mistakes and make suggestions.

jakirkham:
Ah ok, I didn't look too close at what was being tested or how far along it was. I'll try to take a closer look when I have more time. The failure on Travis CI was about installing `apt-get` packages that are normally installed during testing on master so that definitely seems unrelated.

ukoethe:
Replaced with PR https://github.com/ukoethe/vigra/pull/287

PR #290, opened 2016-01-25
Python3 fixes

bluescarni:
This fixes compilation under Linux systems which adopt the new naming scheme for Python libraries (e.g., with trailing ``m``).

There's also a fix for clang compilation (missing ``-pthread`` flag). Does not properly belong to Python 3 fixes, but I figured that with all the ongoing changes in the build system it is easier to push it here.


ukoethe:
manually merged in d151857b8cbf1fa08fdeeea20c5d23036888e3cb

PR #291, opened 2016-01-25
Travis: Switch to container infrastructure

jakirkham:
This should ideally speed up testing on Travis CI. Some packages have been added to a whitelist. All of the packages being installed by `apt-get` are on that whitelist. By using this whitelist, we no longer require `sudo`. By dropping `sudo`, we should be put on the container infrastructure that normally has faster startup times and often fewer users than the regular infrastructure.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,21 @@ language: python
 python:
   - "2.7"
 
+
+sudo: false
+
+addons:
+  apt:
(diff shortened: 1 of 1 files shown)


ukoethe:
Merged into `python3` branch.

jakirkham:
Thanks.

PR #292, opened 2016-01-25
Travis CI Python 3.x fixes

jakirkham:
This appears to do the trick. Turns out this was never picking up the right Python library in the first place.

However, I have found running the following in python points me to the correct directory for the python library.

    from distutils.sysconfig import get_config_var
    get_config_var("LIBDIR")

This doesn't tell me the library itself, but it does tell me where to look. At that point, we can use the existing search method to find the library.

Setting the correct Boost.Python library was simply a matter of searching for `python-pyXY` instead of `python`; where, Python's major version is `X` and minor version is `Y`.

After some more experimentation, it became clear that we didn't have Boost.Python support for 3.5 or 3.4 on Travis CI. Instead, as we have Boost 1.46.1, we only have support for Python 3.2.

Switching to Python 3.2 made a small issue with unicode support apparent. This required the addition of a small bug fix.


jakirkham:
Interestingly, this seems to work everywhere except for Python 2.7 on Travis CI. Seems like it is missing the Python library as a dependency in one place. Not sure why that is. Maybe you could let me know what is going on there, @ukoethe, and we can fix it up.

ukoethe:
The code to configure vigranumpy is in `config/FindVIGRANUMPY_DEPENDENCIES`. I avoid FindPythonLibs there because it was unreliable. Instead, I ask Python directly for all information, and your query of LIBDIR is probably the missing ingredient. (My script already works for Python 2.7 without this addition.) 

Alternatively, I considered `ldd ${PYTHONEXE}` to retrieve the library directly. Would this work on all Linux versions? Can it be done from within Python with a library function instead of a system call?

jakirkham:
> The code to configure vigranumpy is in `config/FindVIGRANUMPY_DEPENDENCIES`.

Yeah, I took a look at it briefly, but wasn't sure I was ready to dig in. It made more sense to me to get a working version going and then clean it up.

> I avoid FindPythonLibs there because it was unreliable.

I agree that the CMake Find modules aren't that great particularly with Python. So, I had also rolled my own, which basically just asks the interpreter about itself and figures out from there. To get things going here, it worked for getting everything set, but we can pull them back out with a little work.

> Instead, I ask Python directly for all information, and your query of LIBDIR is probably the missing ingredient.

I do the same thing. This is what I use for Mac and Linux, but it may work on Windows too. ( https://github.com/nanshe-org/rank_filter/blob/master/cmake/modules/FindPYTHON.cmake ) Basically, I just ripped this code out of there.

> (My script already works for Python 2.7 without this addition.)

So, it did work on Travis, but I think it was picking up the system Python instead of the one Travis was installing in some cases. They install everything in the `/opt` prefix, which you may now realize. There was some PR I did awhile back to try to fix this, but I guess I only fixed *most* of it. The python dynamic library was the straggler. This should fix it.

> Alternatively, I considered `ldd ${PYTHONEXE}` to retrieve the library directly. Would this work on all Linux versions? Can it be done from within Python with a library function instead of a system call?

I suppose, but it feels a little hacky to me. That's why I went for querying the python interpreter directly. As that is what you have done, I think it is better to continue using the interpreter.

On another note, I think I have figured out how to get the right version of Boost.Python.

jakirkham:
It appears that Travis install Boost.Python 1.46.1. This appears to come with Python 2.7 or 3.2 support only.

jakirkham:
So, I switched it to use Python 3.2 and it seems to correctly detect the right Boost library.

jakirkham:
To simplify things, I made a change to `config/FindVIGRANUMPY_DEPENDENCIES.cmake` so that it will search for the Python library in the directory specified by `LIBDIR` first and then by the prefix. This makes sure Travis now picks up the correct python library. As a result, I was able to remove the CMake find strategy and the hacky code that I was doing outside of CMake.

In case you are unclear by what I mean by the wrong Python library, just take a look at this line in the most recent build on `master`. ( https://travis-ci.org/ukoethe/vigra/builds/103574957#L246 ) See how it is from `/usr/lib` instead of `/opt/python/2.7.9`. This was incorrect behavior that seemed to happen before where it might pick up another python library from a standard place. Now this is not happening. ( https://travis-ci.org/ukoethe/vigra/jobs/104784313#L285 )

Given that these changes are doing the correct thing. I think we should include them in `master` as well.

jakirkham:
After making the aforementioned changes, the Python 2.7 tests checked out. The Python 3.x tests appear to still be failing, but likely due to other issues. In short, this appears to be working.

As we now have some good fixes for finding the correct `libpython` and Boost.Python that supports it, I thought it might be best to include those changes in `master` directly as they should avoid some strange building bugs on systems with multiple Pythons. ( https://github.com/ukoethe/vigra/pull/295 ) I think the logic for Mac, Windows, and Linux can be simplified more so that there are no branches, but we can discuss this more.

Given those changes have been broken out. This PR has been transformed to include merging those changes into the `python3` branch and then changing the version of Python 3 here to be compatible with the Boost.Python version provided. It may be worth considering using a newer version of Boost.Python so that we can test against something newer than Python 3.2.

jakirkham:
Appears that the Python 3.x failure is due to the fact that [`PyUnicode_AsUTF8AndSize`]( https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_AsUTF8AndSize ) is not defined on Python 3.2. If we could use a newer version of Boost and similarly a newer version of Python 3 (at least 3.3), this would not present a problem.

jakirkham:
Fixed the Python 3.2 issue in this PR ( https://github.com/ukoethe/vigra/pull/297 ), which has been merged into this PR>

jakirkham:
Updated the summary to more accurately describe what has been done. Barring any major complaints, this appears to be ready to go.

ukoethe:
Superceeded by other PRs.

jakirkham:
Yep. Thanks.

PR #293, opened 2016-01-25
Travis: Include ATLAS to speed up NumPy install

jakirkham:
If NumPy doesn't find a BLAS, it builds a "lite" version of a BLAS itself, which takes awhile. By including a BLAS, we can skip this NumPy build step and see a substantial speed up in the build.


ukoethe:
Merged into `python3` branch.

jakirkham:
Thanks.

PR #295, opened 2016-01-26
CMake, Travis: Fix finding Python library

jakirkham:
This follows up on some earlier work ( https://github.com/ukoethe/vigra/pull/239 ) to make sure Travis CI is using the right Python. Turns out it was still not using the right `libpython` library.

Backports some changes from the `python3` branch in regards to how it finds `libpython`. Includes some additional changes to make sure the best path for `libpython` is used first. Allows pinning the required minor version of Python. Ensures that Boost.Python is used only when it supports the Python interpreter version in use. Specifies the expected Python minor version on Travis.


jakirkham:
See that we are now grabbing `libpython` from the right prefix where Travis installed Python. ( https://travis-ci.org/ukoethe/vigra/builds/104800954#L258 )

This differs from what it was doing previously, which was grabbing the `libpython` from the system Python. ( https://travis-ci.org/ukoethe/vigra/builds/103574957#L246 )

This happens to work here as they are both 2.7 `libpython` and the bug fixes that we were missing didn't hurt us. However, we are grabbing the wrong `libpython`, which will hurt if one has a system with multiple Pythons that have different versions.

ukoethe:
Obsolete.

jakirkham:
Mainly I made this so you could merge it into master without being dependent on the Python 3.x changes. I believe it is already included in `python3` at this point.

ukoethe:
ATM, all development occurs on the `python3` branch, which will be merged into master as soon as possible.

PR #297, opened 2016-01-26
Python 3.2 and lower fixes

jakirkham:
Turns out one minor fix is required for Python 3.2 and lower. This is it.


config/FindVIGRANUMPY_DEPENDENCIES.cmake
```diff
@@ -93,7 +93,7 @@ IF(PYTHONINTERP_FOUND)
     IF(Boost_FOUND)
         FIND_LIBRARY(Boost_PYTHON_LIBRARY
                      NAMES boost_python-py${PYTHON_VERSION_MAJOR}${PYTHON_VERSION_MINOR}   # Linux with multiple Python versions
-                           boost_python-mt                                                 # Mac
+                           boost_python${PYTHON_VERSION_MAJOR}                             # Mac with Python 3
```
jakirkham: I believe `mt` is just for multithreading. I suppose we could add it to the list. Though it should be named differently namely add `boost_python${PYTHON_VERSION_MAJOR}-mt and then `boost_python${PYTHON_VERSION_MAJOR}-mt`.

ukoethe: The convention `boost_python-mt` actually applies to our lab's test Mac, so it cannot be removed. Is there a system where `boost_python${PYTHON_VERSION_MAJOR}` is used, or is this just your suggestion of how the library should be named?

ukoethe:
Merged into `python3` branch.

config/FindVIGRANUMPY_DEPENDENCIES.cmake
jakirkham: Yes, this happened in the CI. We can keep `mt`, but I suggest we move it down the list. We should prefer the non-`mt` version as we have done in other cases here first for consistency.

jakirkham: I've updated this to include multithreaded libraries as a last resort. In all cases, I have tried to get the correction Python version, if possible.

config/FindVIGRANUMPY_DEPENDENCIES.cmake
```diff
@@ -92,9 +92,12 @@ IF(PYTHONINTERP_FOUND)
...
         FIND_LIBRARY(Boost_PYTHON_LIBRARY
-                     NAMES boost_python-py${PYTHON_VERSION_MAJOR}${PYTHON_VERSION_MINOR}   # Linux with multiple Python versions
-                           boost_python-mt                                                 # Mac
-                           boost_python                                                    # default
+                     NAMES boost_python-py${PYTHON_VERSION_MAJOR}${PYTHON_VERSION_MINOR}      # Linux with multiple Python versions
+                           boost_python-py${PYTHON_VERSION_MAJOR}${PYTHON_VERSION_MINOR}-mt   # Linux with multiple Python versions (multithreaded)
+                           boost_python${PYTHON_VERSION_MAJOR}                                # Mac with Python 3
+                           boost_python${PYTHON_VERSION_MAJOR}-mt                             # Mac with Python 3 (multithreaded)
+                           boost_python                                                       # default
+                           boost_python-mt                                                    # Mac with Python (multithreaded)
```
jakirkham: Added this one back for Mac when a single threaded version cannot be found.

jakirkham:
Resolved conflicts. Also, added multithreading libraries for all platforms.

config/FindVIGRANUMPY_DEPENDENCIES.cmake
ukoethe: Why did you place the `-mt` variants after the ones without `-mt`? Intuitively, I'd have done it the other way around.

jakirkham: Perhaps. I am not sure I really follow what is getting multithreaded with the Python bindings. Do you know what is going on here?

Well, now that I think about it, I think we should be respecting `Boost_USE_MULTITHREADED`. ( https://cmake.org/cmake/help/v3.0/module/FindBoost.html ) That way we add the libraries by default, but if it is turned off we don't include them.

ukoethe: The cmake documentation says:
```
Boost_USE_MULTITHREADED  - Set to OFF to use the non-multithreaded
                           libraries ('mt' tag).  Default is ON.
```
This looks like a mistake. I think the non-multithreaded libraries do *not* have the 'mt' tag (the converse is not true: if the library name doesn't contain the 'mt' tag, it can still be multi-threaded). 'Multi-threaded' here actually means 'thread-safe' (see http://www.boost.org/build/tutorial.html), so one would always prefer the 'mt' tag when the program might use multiple threads.

jakirkham: So it may not be doing any threading, but it is placing locks around any global state manipulations or using atomics.

Ok, we can place them first, but I still think the `Boost_USE_MULTITHREADED` variable should be respected.

ukoethe: Ok, I pushed a solution in 939bdee2907e1359466184bc7f53af44d703bbca. Please check.

jakirkham: Cool. I think it looks good.

jakirkham:
I think you will still want this commit ( 5b7a5cc6412b615eacaff2894d4dc3b5353005f7 ) as there were some remaining references to `VIGRANUMPY_THREAD_LIBRARIES`, which should be replaced as it is no longer defined. Also, it looks like FFTW requires threading now so I added this commit ( c8a6d8b9cac5db72f85c108e84c6da4c68553771 ). The rest appeared to be resolved in a better or the same way in your work.

ukoethe:
`VIGRANUMPY_THREAD_LIBRARIES` is still defined, I just moved it one directory level up :-). I fixed `fourier` in 356444be03d355e2b893de8c1288866abc59bad2.

Thanks for your input.

jakirkham:
I see it now. Maybe I was missing that commit before.

PR #299, opened 2016-01-26
Travis for Python 3

ukoethe:


config/FindVIGRANUMPY_DEPENDENCIES.cmake
```diff
@@ -2,65 +2,108 @@
...
+            CACHE FILEPATH "Python libraries"
+            FORCE)
+    ELSE()
+        IF(WIN32)
+            set(PYTHON_LIBRARY_NAME python${PYTHON_VERSION_MAJOR}${PYTHON_VERSION_MINOR})
+        ELSE()
+            execute_process(COMMAND ${PYTHON_EXECUTABLE} -c
+                             "from distutils.sysconfig import *; print(get_config_var('LDLIBRARY'))"
+                              OUTPUT_VARIABLE PYTHON_LIBRARY_NAME OUTPUT_STRIP_TRAILING_WHITESPACE)
+            execute_process(COMMAND ${PYTHON_EXECUTABLE} -c
```
jakirkham: Before this was run on Windows and Linux, which worked for me. However, now it appears this is just being run on Linux. I think this is what is causing AppVeyor to fail.

config/FindVIGRANUMPY_DEPENDENCIES.cmake
```diff
@@ -2,65 +2,108 @@
...
+    #
+    #      find Python library
+    #
+    ######################################################################
+    IF(APPLE AND ${PYTHON_PREFIX} MATCHES ".*framework.*")
+        SET(PYTHON_LIBRARIES "${PYTHON_PREFIX}/Python"
+            CACHE FILEPATH "Python libraries"
+            FORCE)
+    ELSE()
+        IF(WIN32)
```
jakirkham: Regarding Windows, you know more than I do, but should something be done for Win64, as well? Would this work for both?

jakirkham:
How have you configured AppVeyor? I was going to run a git bisect to see what went wrong, but I don't seem to have an `appveyor.yml` file to test with.

config/FindVIGRANUMPY_DEPENDENCIES.cmake
ukoethe: According to my understanding of the cmake documentation (https://cmake.org/Wiki/CMake_Useful_Variables), WIN32 applies to all Windows versions.

jakirkham: Ah ok. Sorry, I was confused.

ukoethe:
I think I configured AppVeyor directly on their website. It's much easier than the debug-by-pushing cycle of travis, but at the price that you can't see or edit it..

ukoethe:
Shall we add Mac builds to the travis matrix as well?

jakirkham:
Yeah, I like being able to have my CI configs in revision control. I made the less desirable choice of opening PRs against this to test instead.

Adding Mac builds sounds good.

config/FindVIGRANUMPY_DEPENDENCIES.cmake
jakirkham: Turns out this is not the case and does nothing on Windows.

jakirkham:
I am working on a Mac PR ( https://github.com/ukoethe/vigra/pull/303 ). It is not ready yet, but I will let you know when it is.

jakirkham:
It might also be worth considering using the Ubuntu Trusty environment, as well. This would provide access to a newer version of Boost and hopefully allow testing against newer versions of Python.

ukoethe:
Merged into `python3` branch.

jakirkham:
Thanks.

PR #300, opened 2016-01-26
Travis for Python 3 Fixes (pt. 2)

jakirkham:
* Search for `LIBDIR` on Windows and Linux.
* Ensure the right Python component library is used.
* Bring back Travis CI testing.


jakirkham:
Maybe this isn't the solution. Not sure what is going wrong with AppVeyor here. Look at a diff between this and the original PR ( https://github.com/ukoethe/vigra/pull/292 ) doesn't show anything obviously wrong.

CMakeLists.txt
```diff
@@ -134,7 +134,7 @@ ENDIF()
 
 IF(WITH_VIGRANUMPY)
     set(WITH_BOOST "ON")
-    # set(WITH_BOOST_COMPONENTS ${WITH_BOOST_COMPONENTS} "python-py${PYTHON_VERSION_MAJOR}${PYTHON_VERSION_MINOR}")
+    set(WITH_BOOST_COMPONENTS ${WITH_BOOST_COMPONENTS} "python-py${PYTHON_VERSION_MAJOR}${PYTHON_VERSION_MINOR}")
```
ukoethe: I deliberately don't add `python` as a boost component because I found that the `FIND_LIBRARY` call in `FindVIGRANUMPY_DEPENDENCIES.cmake` is more reliable.

config/FindVIGRANUMPY_DEPENDENCIES.cmake
```diff
@@ -68,10 +68,10 @@ IF(PYTHONINTERP_FOUND)
...
                              "from distutils.sysconfig import *; print(get_config_var('LDLIBRARY'))"
                               OUTPUT_VARIABLE PYTHON_LIBRARY_NAME OUTPUT_STRIP_TRAILING_WHITESPACE)
-            execute_process(COMMAND ${PYTHON_EXECUTABLE} -c
-                             "from distutils.sysconfig import *; print(get_config_var('LIBDIR'))"
-        			           OUTPUT_VARIABLE PYTHON_LIBRARY_PREFIX OUTPUT_STRIP_TRAILING_WHITESPACE)
         ENDIF()
+        execute_process(COMMAND ${PYTHON_EXECUTABLE} -c
+                        "from distutils.sysconfig import *; print(get_config_var('LIBDIR'))"
+                        OUTPUT_VARIABLE PYTHON_LIBRARY_PREFIX OUTPUT_STRIP_TRAILING_WHITESPACE)
         FIND_LIBRARY(PYTHON_LIBRARIES ${PYTHON_LIBRARY_NAME} HINTS "${PYTHON_LIBRARY_PREFIX}" "${PYTHON_PREFIX}"
```
ukoethe: Doing this on Windows has not effect since `get_config_var('LIBDIR')` returns `None`. However, the Windows build works for me, so don't worry about it. (The AppVeyor failure was do to `FIND_PACKAGE(PythonInterp ${PYTHON_VERSION})` being called with empty PYTHON_VERSION variable. I simply had to remove the quotes.)

CMakeLists.txt
jakirkham: I don't follow how that relates to adding Boost.Python. In the current strategy, we are not looking for Boost.Python component and not guaranteeing it was build against the right version of Python. I have found that not setting this will cause us to pick up the wrong Boost.Python instead.

By making this change, I was able to discover that we were picking up the wrong Boost.Python (namely 3.2 when trying to test 3.4 or 3.5). Once added it would simply skip testing on Python with versions 3.4 or 3.5 because it couldn't find a compatible copy of Boost.Python. All of this was possible because this constraint ensures only a copy Boost.Python that works with our interpreter is picked.

config/FindVIGRANUMPY_DEPENDENCIES.cmake
jakirkham: Oh, I see. Weird that it wouldn't just ignore the empty quotes.

CMakeLists.txt
ukoethe: In theory, you are right that this is the best approach. I tried it before resorting to `FIND_LIBRARY` and found it to be non-portable, because the `python-py*` naming convention is not universally followed (I forgot who caused problems).

jakirkham: Could we just add branches for different platforms here? I am a little concerned that we get something we don't like by not pinning this somehow.

ukoethe: `FIND_LIBRARY` is configured such that it will prefer the `python-py*` variant if it exists, but falls back to alternative conventions otherwise. It can indeed happen that it picks an incompatible boost_python version this way, but the WITH_BOOST_COMPONENTS approach just fails the other way around - it may miss a perfectly valid library.

jakirkham:
I think this is no longer necessary so closing.

PR #303, opened 2016-01-26
Travis Mac Support

jakirkham:
Adds Travis CI support for Mac.

All that still needs to be done is to fix a segmentation fault that is occurring in `test_classifier`.


jakirkham:
Kind of stuck ATM. Appear to be hitting this bug ( https://llvm.org/bugs/show_bug.cgi?id=21083 ). It occurs the same way with octave and some other things.

1. We could try installing a different compiler than the system one with brew.
2. Could try to get a new copy of XCode.

On point 2, if there is a version of XCode that fixes this we could ask Travis to upgrade, but I don't know if a newer version with the fix would support Mavericks (meaning they might need to update their VMs OS version). So, that would have a rather slow turnaround.

Seems worthwhile to try 1.

ukoethe:
Do you have the full error message? It might be easiest to implement a workaround in vigra.

ukoethe:
Of you disabled Linux for the moment, Travis might respond faster.

ukoethe:
Yes, vigra sets '-std=c++11' automatically to get std:: thread.

jakirkham:
> Do you have the full error message?

Sure. ( https://travis-ci.org/jakirkham/vigra/jobs/105000389#L2943-L2946 ).

> It might be easiest to implement a workaround in vigra.

Doubtful, appears to be an issue with the headers shipped with clang.

> Of you disabled Linux for the moment, Travis might respond faster.

Maybe. I think they run on different architectures so they have different queues and wait times. The Linux builds start right away now as they are in containers. I saw a listing of startup times somewhere and Mac is one of if not the longest.

> Yes, vigra sets '-std=c++11' automatically to get std:: thread.

Yeah, I forgot until I saw the message from cmake.

<br>
So, this is what I have tried thus far.

1. System compiler.
2. New copy of llvm/clang from homebrew.
3. Old apple gcc-4.2 from homebrew.

These are all bust. Here is why.

System compiler ends up having an llvm bug that will probably be difficult to workaround and is fixed in a later version (keep in mind we are talking about 1 yr ago). New copy of XCode may fix it if it has the working copy of clang. I don't know that we can update XCode ourselves so this would require convincing Travis to upgrade it.

Homebrew failed to find a bottle (binary) for llvm/clang. Maybe messing around with the options might change this. As homebrew didn't find a bottle, it had to build from source. The time scale for this is simply unreasonable in the CI. Also, when trying to wait it out, the build failed with an exit code (137) than suggests we were running out of resources and Travis killed it. So, we couldn't build it if we wanted to.

Turns out the original Mac gcc compiler is so old that we can't build the parts we want to test like vigranumpy. Also, it seemed to pick up the includes that the system compiler was using so it got pretty messed up. Trying to fix this appears to be a chore and is not really worthwhile.

The only thing worth trying at this point would be to install a reasonably new version of gcc (4.8 or 4.9) or find a bottled version of clang. I can't seem to find the list of available bottles and checking it via brute force with Travis is super slow. Beyond that I am out of ideas.

If someone else has ideas or more time, I think this PR would be a good starting point. It is at a point where it appears to be setting up everything correctly, but is merely failing during the build of VIGRA. It would be possible for someone to pick this up and work on it at this point even if the don't have a Mac.

jakirkham:
So, I have tried a bunch of things. I think we can be fairly confident of how Boost.Python is built now. Unfortunately, it picks up the system Python with 2.7. To fix this, we will have to try something pretty destructing in the VM. For example, we could replace the system interpreter with a link to our Python or we could fix the linkage on Boost.Python libraries afterwards. However, both are Python 2.7 and it doesn't seem like this is causing any problems at present. Boost.Python correctly links against Python 3.4.1.

Switching to the gcc compiler has improved the build. Also, using Boost Threads was required for gcc on Mac. It appears that the C++11 requirement is becoming more necessary so this is explicitly enabled on Mac.

Still we seem to be having issues during the build. It appears that it could be due a mismatch in architectures for some libraries. In particular, this seems to occur with Boost.Python. It may also involve the Python library.

jakirkham:
Well, I have tried to get this to work and it seems like it is very close, but it is not working. Honestly, I don't have any more time to put into this. Maybe you can ask @stuarteberg for some help. I know he builds VIGRA with a more recent gcc compiler so he might have some ideas about things I have missed.

ukoethe:
We are just looking into how much work it would be to make the threadpool compatible with boost:thread. This would at least solve one problem.

>  I don't have any more time to put into this.

Thanks for you time anyway. We can always continue on this PR later.

jakirkham:
> We are just looking into how much work it would be to make the threadpool compatible with boost:thread. This would at least solve one problem.

I see. Well, at least we are to a point where we can consider what needs to be done for the Mac build.

> Thanks for you time anyway. We can always continue on this PR later.

Sure. Sorry, I couldn't get it to work. If you need any help understanding what happened here, please let me know. I'd be happy to explain.

jakirkham:
Had a little more time to look at this. Turns out the issue with missing Boost.Python symbols did resolve down to differences between libc++ and stdlibc++. As VIGRA can't be built with clang due to the aforementioned clang bug, we build it with gcc. By persuading brew to build Boost.Python with gcc, this issue largely went away.

There were a few other linking errors, but are much simpler to diagnose. For instance, Boost.System is needed by analysis. Also, it appears the FFTW bindings depend on Boost threading. At this point, I am waiting for this last one to test out.

jakirkham:
Good news, the build completes. Also, the vigranumpy tests pass. :tada:

The VIGRA test suite is giving me two segmentation faults in two tests. This happens for both Python 2 and 3, which should come as no surprise. The failing tests are `test_classifier` and `test_multiarray_chunked`. It doesn't give me lots of info, but this is what it says on the Python 2 build. The Python 3 build is the same except the numbers following the line numbers and preceding the Segmentation fault differ.

          Start  5: test_classifier
     5/52 Test  #5: test_classifier ..................***Failed    0.99 sec
    /Users/travis/build/ukoethe/vigra/build/test/classifier/run_test_classifier.sh: line 1: 30609 Segmentation fault: 11  ./test_classifier
          Start 29: test_multiarray_chunked
     29/52 Test #29: test_multiarray_chunked ..........***Failed    1.15 sec
    /Users/travis/build/ukoethe/vigra/build/test/multiarray/run_test_multiarray_chunked.sh: line 1: 30686 Segmentation fault: 11  ./test_multiarray_chunked

Unfortunately, I can't give any more help than that. Alright, now, I really need to leave this.

jakirkham:
Tried to clean up homebrew cache, which eliminated ~0.33GB, but it didn't eliminate the seg fault. My best guess is maybe there is still some non-Boost Thread and Boost Thread intermixing that is causing something bad to happen.

jakirkham:
I guess I forgot to say this in a comment and left it in a commit message ( https://github.com/jakirkham/vigra/commit/08a68b3ac86dade3d34e4b082b40419e9aa1bb0a ) instead. To make sure I (and anyone else that encounters this problem) remembers, I am going to write it here. The reason we could not use gcc c++11 features for threading here is that gcc on Mac is that we cannot use `std::timed_mutex` or `std::recursive_timed_mutex`. This is because gcc do not have access to a library function needed to implement this. Here is the most relevant gcc bug report that summarizes why they don't support this best. ( https://gcc.gnu.org/bugzilla/show_bug.cgi?id=50196#c8 )

I don't know if using Boost will be a viable to workaround for this issue, but if it is that would be a way to fix it (what I tried here). Alternatively, getting clang support back for Mac and switching back to that compiler would be another way to fix it (however that seems unlikely unless we get a copy of clang that has the fix for this https://llvm.org/bugs/show_bug.cgi?id=21083 or we use Boost for these sorts of resolutions instead). Perhaps even simpler than either of these would be just to drop the use of `timed_mutex` and `recursive_timed_mutex` as I think we merely include them in `threading.hxx` and never actually use them anywhere.

PS - @stuarteberg, sorry I incorrectly remembered the reason earlier.

jakirkham:
I added this commit ( https://github.com/jakirkham/vigra/commit/67f06ccb00d217d24655bf3cef7a657efa4f5f20 ) to workaround the issues with gcc and timed mutexes. It appears to work and I think it would be worthwhile to cherry pick it where needed. I know it works because I was able to turn off Boost Threads and it built all the way through. This also appears to eliminate one of the segmentation faults in my CI build of it. In particular, it eliminated the `test_multiarray_chunked` failure. Now only `test_classifier` fails with a segmentation fault.

jakirkham:
This PR now contains homebrew bottles to speed up the Boost Python step.

jakirkham:
I have opened this issue with Travis CI ( https://github.com/travis-ci/travis-ci/issues/5529 ) regarding getting a newer copy of XCode on the VMs.

jakirkham:
Turns out we can select a newer version of XCode. I tried getting the latest version of 6, but it didn't work. So, I grabbed the latest version of 7 and that seems to work. Dropped python environment selection, the old bottles, and gcc. Everything appears to work with the system compiler (clang).

jakirkham:
Closing in favor of the cleaned up version of this PR ( https://github.com/ukoethe/vigra/pull/308 ).

PR #304, opened 2016-01-27
Clang fixes

bluescarni:
A couple of problems spotted after compilation with clang, and subsequent run through the thread sanitizer.


ukoethe:
Merged into `python3` branch.

PR #309, opened 2016-01-29
Travis Linux Python 3.4 (old)

jakirkham:
Currently we have Python 3.2 testing on Linux, but it is a bit of an ancient version of Python 3. This adds testing for the more recent Python 3.4 on Linux. It may be worth discussing if we want to keep the Python 3.2 support or drop it in favor this.


ukoethe:
When I tried Travis with Python 3.4 a week ago, it didn't work out of the box. I was glad that you found out that Python 3.2 would work. So, if your new configuration works, I don't mind upgrading the test.

jakirkham:
Yeah, the trick is Ubuntu Precise only has a copy of Boost Python for 3.2. However, they let us use Ubuntu Trusty, which has Boost Python for 3.4. The only downsides to dropping Precise/3.2 testing would be Trusty does not run on the container infrastructure so is slower to startup and comes from a different (probably more limited) queue and testing oddities that we have seen for 3.2. Though this last point only matters if we are really interested in testing a standard Ubuntu Precise install.

jakirkham:
Hmm...seems Travis CI is building Python with a static library. Opened this issue to try and get it built with a shared library ( https://github.com/travis-ci/travis-ci/issues/5533 ).

ukoethe:
Yes, let's wait until the Travis folks fix this.

jakirkham:
Sounds good. Plus, we now have Windows tests on 3.5.1, as well.

jakirkham:
Replaced by this PR ( https://github.com/ukoethe/vigra/pull/321 ).

PR #311, opened 2016-02-01
AppVeyor Python 2/3 build matrix

jakirkham:
Adds support for testing on both Python 2 and 3 with AppVeyor.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,60 @@
+matrix:
+  include:
+    - os: linux
+      sudo: false
+      language: python
+      python: "2.7"
+    - os: linux
+      sudo: false
+      language: python
+      python: "3.2"
+    - os: osx
+      osx_image: xcode7.3
+      language: generic
+      env: TRAVIS_PYTHON_VERSION="2"
+    - os: osx
+      osx_image: xcode7.3
+      language: generic
+      env: TRAVIS_PYTHON_VERSION="3"
+
+
+addons:
+  apt:
+    packages:
+      - libhdf5-serial-dev
+      - libboost-python-dev
+      - libjpeg-dev
+      - libtiff4-dev
+      - libpng12-dev
+      - libfftw3-dev
+      - libatlas-base-dev
+      - cmake
+
+before_script:
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then export PATH="/usr/local/bin:${PATH}"                                                ; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew remove --force $(brew list); brew cleanup --force                              ; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then travis_retry brew update                                                            ; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew tap homebrew/science                                                           ; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install cmake                                                                  ; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install hdf5                                                                   ; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install boost                                                                  ; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install libjpeg                                                                ; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install libtiff                                                                ; fi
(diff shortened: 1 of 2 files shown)


jakirkham:
Interesting, doesn't seem to work with Python 2.x. One thought is we need an older version of VC++. This should be possible with AppVeyor. However they don't actually support VC++ 2009. Instead they support VC++ 2008 and VC++ 2010. Trying these might help the problem.

ukoethe:
The oldest compiler version to be supported by the upcoming release is Visual Studio 2012 (vc11). If AppVeyor provides this version, you should be able to get a working build by
```
    - conda create -n python -c ukoethe --override-channels visual-studio=11.0 python=2.7 jpeg libpng libtiff hdf5 numpy nose boost fftw
    - activate python
    - cd %APPVEYOR_BUILD_FOLDER%
    - mkdir build
    - cd build
    - cmake .. -G "Visual Studio 11 Win64" -DVIGRANUMPY_REQUIRED=1 -DDEPENDENCY_SEARCH_PREFIX="%CONDA_ACTIVE_ENV%/Library"
    - cmake --build . --target check --config Release
```
(These are the dependencies I use for building ilastik on Windows.)

jakirkham:
Yeah, Stuart mentioned to me that the Visual Studio situation was complicated by the fact that some C++0x features are required, but CPython is only supported to build on one version of VC++, which is too old for these features. Thanks for telling me which version is required.

They do support this version of Visual Studio.

I have written a patch to include these changes. If it works out on my temporary branch, I will merge it into here.

Does it really require the `--override-channels` flag? I would think that by specifying the version of VC++ this just works. What problems do you encounter when this isn't used?

ukoethe:
> Does it really require the `--override-channels` flag?

I didn't check this for the specific conda call here, but in general conda repeatedly preferred packages from the default channels when it shouldn't. This might be a consequence of the limitations of conda's present version resolution algorithm, but I found it safest to switch off the default channels altogether.

jakirkham:
Fair enough. I was hoping to keep the code simple here, but I don't want to do it at the expensive of it not behaving correctly.

This appears to work on Python 2 in my temporary branch. Am still waiting for the Python 3 part to complete, but expect no surprises there. Anyways figured it was time to update what is present here.

jakirkham:
So, this may have been accidental, but I think this got closed when you closed the AppVeyor PR. Would you like me to reopen this against the `python` branch or was this intentionally closed for some reason?

ukoethe:
I'll create a new PR in a second, with your changes cherrypicked.

ukoethe:
I checked the appveyor output of your py27 build to figure out what happend. Hopefully, it will now succeed, see https://github.com/ukoethe/vigra/pull/312.

jakirkham:
Yeah, so this build is basically the same as the one here except I remove the `.travis.yml` file again and it passed both Python 2 and 3. ( https://ci.appveyor.com/project/jakirkham/vigra/build/1.0.13 ) I already am running the AppVeyor with the changes here, but it is not done yet. ( https://ci.appveyor.com/project/jakirkham/vigra/build/1.0.14 )

ukoethe:
Unfortunately, it passed the Python 2 test only accidentally - it didn't actually execute the vigranumpy tests (`nose` didn't work), and my attempt to enforce this via `VIGRANUMPY_REQUIRED=1` was too weak. Trying to figure out the problem with nose ...

jakirkham:
Ah ok. Please let me know if I can help. This seems like the last CI test we really should have.

ukoethe:
> This seems like the last CI test we really should have. 

Yeah, then the Python3 PR is ready for merge.

jakirkham:
So, there is `nose`, but not one with a feature for VC++. It appears to downgrade it 1.2.1. I am guessing this is too low of a version so it is not detected. Though this does not happen with Python 3. It only happens with Python 2.

Not sure why that is. I think this is just another case of features being weird in `conda` as I have seen similar behavior before Maybe you could trying running `conda update nose` after the install to see if this will fix it. I don't think anything else is required.

ukoethe:
It seems that nose 1.2.1 is broken in channel `ukoethe`. I just uploaded nose 1.3.7 which I had been using locally for a while, but forgot to upload. A feature is not needed because nose is pure Python. Still waiting for AppVeyor to start...

jakirkham:
Ah ok. Nevermind. Must be confused.

jakirkham:
I restarted AppVeyor and it seems to be picking up and detecting `nose` is there. It is also no longer being downgraded.

jakirkham:
This is now passing for me on Python 2 and I am seeing that it is running the vigranumpy tests. ( https://ci.appveyor.com/project/ukoethe/vigra/build/1.0.444/job/a096p6avb7wucu7i )

PR #315, opened 2016-02-02
Boost thread pool

bluescarni:

PR #316, opened 2016-02-02
Boost thread pool

ukoethe:
Fixed PR https://github.com/ukoethe/vigra/pull/315 on Windows


ukoethe:
The problem was that `BOOST_THREAD_PROVIDES_VARIADIC_THREAD` is not defined for Visual Studio (even for VC 14). Thus, `packaged_task` only supports functions without arguments. I fixed this by binding the thread ID parameter to the fixed value 0, which happens to work for VIGRA's current use cases of `parallel_for`, but is of course dangerous in general. @bluescarni Are there any better solutions that don't need a lot of work?

ukoethe:
Pinging @jakirkham : The travis test with `WITH_BOOST_THREAD=1` failed because the required boost libraries were not found. Do you have any idea what happened?

jakirkham:
Not sure. Maybe change `libboost-python-dev` to `libboost-dev`? It might just be it isn't installing the other needed Boost libraries. At least it is a good place to start.

jakirkham:
Scratch that. I think there was an issue with the version of the Boost Libraries provided in precise. In particular, they were missing a few things that were included in later versions. Not sure what they were though.

jakirkham:
Yep, there is no `libboost_chrono`. That is why it is failing. I don't know how hard of a dependency that is for you or whether it is worth trying to remove.

jakirkham:
Appears it got added to the next minor release of Boost after the one included with Precise. Namely 1.47.0. ( http://www.boost.org/users/history/version_1_47_0.html )

jakirkham:
Though Travis CI added Boost 1.55 to the whitelist. ( https://github.com/travis-ci/apt-package-whitelist/blob/master/ubuntu-precise#L4071 )

I believe you can add the `apt-get` list to the matrix, but honestly switching to 1.55.0 might not be a big deal for all the Linux tests.

ukoethe:
It seems that boost_chrono is not needed with boost 1.54 on Linux.

Anyway, @stuarteberg doesn't need the option `WITH_BOOST_THREAD` anymore. I tend to the decision to drop it completely.

jakirkham:
Yeah, I don't use it either. I just build with clang and use C++11 flags.

ukoethe:
So I think I'll give up on this. See PR https://github.com/ukoethe/vigra/pull/317.

jakirkham:
Yeah, sounds smart. Ultimately cutting down on these weird corner cases where C++11 solves the problem seems like the right move.

jakirkham:
Not sure if it matters any more, but I think I fixed it. See this PR ( https://github.com/ukoethe/vigra/pull/318 ). I don't if you will accept it, but I think it would be nice to have at least one commit on `master` where this is all working before tearing out Boost.Thread just in case we find we need it for some reason.

bluescarni:
@ukoethe I would have hoped that in absence of variadic templates ``packaged_task`` would use some kind of Boost macro trickery to emulate them, but I just checked and apparently it is not the case. I don't see much a way forward in this sense, so maybe we can keep ``BOOST_THREAD`` with caveats for the next version and finally drop it after the next release.

PR #319, opened 2016-02-03
Python 3 Mac bottle

jakirkham:
Adds a homebrew bottle (binary) for Boost.Python built for Python 3. This is used to speedup the Travis CI tests for Mac Python 3 by a few minutes (~6 mins). It will make the Python 3 and Python 2 tests on Mac take a comparable amount of time by skipping the otherwise necessary build step for Boost.Python on Python 3. This may also help avoid the infrequent hanging that sometimes occurs with the Mac Python 3 build. The bottle only takes up 6.7MB.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,7 @@ before_script:
   - pip install nose
   - pip install numpy
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]] && [[ ${TRAVIS_PYTHON_VERSION:0:1} -eq 2 ]]; then brew install boost-python --with-python ; fi
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]] && [[ ${TRAVIS_PYTHON_VERSION:0:1} -eq 3 ]]; then brew install boost-python --without-python --with-python3 ; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]] && [[ ${TRAVIS_PYTHON_VERSION:0:1} -eq 3 ]]; then brew install "mac_homebrew_bottles/brew-python/${TRAVIS_PYTHON_VERSION}/boost-python-1.60.0.el_capitan.bottle.1.tar.gz" || brew install boost-python --without-python --with-python3 ; fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew cleanup --force -s; sudo rm -rf $(brew --cache)                                     ; fi
 
 script:


jakirkham:
Seems to work fine. If you want to put the homebrew bottle somewhere else, you could put it on [bintray.com]( https://bintray.com/ ).

ukoethe:
Given that we eventually want to move to conda anyway, I'm not sure if this change is worthwhile. If you think it is, please upload the bottle to bintray or another public site and create an updated PR. I don't want the bottle to be part of the repo.

jakirkham:
> I'm not sure if this change is worthwhile.

I just kept running into issues with the Mac Python 3 build hanging and so figured that it is running into some resource issues. This would help with that. Alternatively, one could just restart this Mac Python 3 build when that happens.

> we eventually want to move to conda anyway

Yeah, I expect this will only be useful until we get there and then be removed.

> If you think it is, please upload the bottle to bintray or another public site and create an updated PR.

I thought you might want access to whatever site it would be uploaded to so figured I wouldn't upload it unless I would have a way to transfer it. Plus, considering the Windows dependencies were here, I figured it wouldn't be that big of a deal if one Mac dependency was here especially given it is comparable in size.

FWIW, I got this bottle from an existing bintray link so I suppose that link could be used, but it was discouraged as a permanent solution. ( https://github.com/Homebrew/homebrew/issues/47218#issuecomment-177844653 )

> I don't want the bottle to be part of the repo.

Understood. Please let me know what you think about the stuff above.

ukoethe:
> I thought you might want access to whatever site it would be uploaded

I suppose that everyone needs read access in order for the tests to execute. So I can just download it when I need to. I don't need write access for now.

> considering the Windows dependencies were here

Will be removed in the release.

jakirkham:
This may soon be unnecessary given there is a plan to build the recent version of boost. ( https://github.com/conda/conda-recipes/issues/579 )

The only other package that is not already supplied by the `default` channel is `fftw`. I have built a copy for Mac and Linux, which you are welcome to copy. Though I am sure there is an ilastik channel with this package.

jakirkham:
So, I just built the latest recipes for Python 2.7 and 3.5 and put them in my own channel (same name as my username if you want to copy or use them). Maybe we should just focus on getting working conda based builds. What do you think?

In any event, the current strategy works and I think that putting any more effort into this quick workaround is a waste. So, I will close it.

PR #322, opened 2016-02-04
Version 1.11.rc

ukoethe:
Needs a bit more testing and some documentation cleanup.


jakirkham:
@stuarteberg, could you try building this and see if it works for you? I will also do the same.

jakirkham:
So, I managed to build this on Mac and Linux. Though I had to change some of my build flags to get it to go through.

On Mac I built Boost and VIGRA with C++11 support and disabled Boost.Thread. On Linux, I had to build with Boost.Thread even though I had C++11 support.

This last part feels a bit weird to me. So, I will need to investigate more as to why this was the case.

ukoethe:
> On Linux, I had to build with Boost.Thread even though I had C++11 support.

This is indeed strange, as it works on our local machines and on Travis.

jakirkham:
Yeah, it is weird. There are three main difference I can think of.

* Using VirtualBox
* CentOS 6.3
* Non-standard system compiler

There are some weird cases with VirtualBox where it disallows some types of instructions like AVX/AVX2, but I don't know of any threading constraints though there may be some. I'm not sure why the operating system matters here. The compiler I am using is still supplied by a CentOS repo, but it is supplied as an add-on and is a newer version (GCC 4.8.2). I believe this compiler is similar if not the same as the add-on one used on the Janelia cluster; hence, why it was chosen.

jakirkham:
FWIW, I did try building against a beta of NumPy 1.11.0 and it went ok. So, hopefully switching to that version should be relatively straightforward.

ukoethe:
@jakirkham : after merging your changes to `.travis.yml`, the OS X tests still took 48 minutes and timed out. Can you have a look what happened? Thank you.

jakirkham:
Looks like the failure is actually occurring on Mac Python 2 tests, which previously did not have a problem. These tests would not have been impacted by my change.

Both Mac Python 2 and 3 tests appear to be taking the same length of time ~48min give or take a couple of seconds. Not sure why 5secs pushes the Mac Python 2 test over.

I do find it concerning that the total test time is now ~48min as opposed to ~43min. Naively, I would say this is due @stuarteberg's change. Though I can't figure out why that would result in ~5min more. Definitely requires more investigation.

jakirkham:
Looks like one of the Linux tests died. ( https://travis-ci.org/ukoethe/vigra/jobs/111290319#L946 )

jakirkham:
Maybe we need to start cleaning up some build intermediates before testing. This may be hurting us on our memory limits.

hmeine:
Do you need additional tests on Linux (w.r.t. c++11 / boost.threads)?

jakirkham:
:tada:

PR #323, opened 2016-02-05
fix typo in check which makes gaussianDerivative() unusable from python

svenpeter42:
fix typo in check which makes gaussianDerivative() unusable from python


ukoethe:
resolved in ed3a62b901d3c9a3723bdf7810b695f820465fa8

PR #324, opened 2016-02-09
Multideftests

dagophil:
Added vigranumpy tests for functions that were exported via multidef.

### DIFF
--- a/vigranumpy/lib/__init__.py
+++ b/vigranumpy/lib/__init__.py
@@ -633,32 +633,6 @@ _genTensorConvenienceFunctions()
 del _genTensorConvenienceFunctions
 
 
-
-
-
-# define tensor convenience functions
-def _genDistanceTransformFunctions():
-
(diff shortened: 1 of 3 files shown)


ukoethe:
What happened to the distance transform problem we talked about?

dagophil:
I planned to add the commits for the distance transform fix in a separate pull request. Should I instead put them here?

ukoethe:
Yes, just add them here, the PR will be updated automatically.

dagophil:
About c060eef: The function `vigra.filters.distanceTransform` was defined multiple times. The new definition in `morphology.cxx` was ignored because of the old definition in `__init__.py`. I just threw out the old definition, since it is not needed anymore.

ukoethe:
Now the test doesn't work anymore...

ukoethe:
Merged into version-1-11-rc branch.

PR #328, opened 2016-02-19
Fix signature of PythonRegionFeatureAccumulator.remappingMerge()

stuarteberg:
This change seems to be necessary for `RegionFeatureAccumulator.merge()` to work in python. (Otherwise it complains that there is no matching C++ function to call.)


ukoethe:
merged into https://github.com/ukoethe/vigra/tree/version-1-11-rc

PR #329, opened 2016-02-22
WIP: New Edge Feature Accumulation

consti123:
Added function to get more statistics over the RAG Edges, relying on Boost accumulators.

### DIFF
--- a/vigranumpy/lib/__init__.py
+++ b/vigranumpy/lib/__init__.py
@@ -1223,12 +1223,12 @@ def _genRegionAdjacencyGraphConvenienceFunctions():
                 else:
                     return graphs._ragEdgeFeatures(self,graph,affiliatedEdges,edgeFeatures,weights,acc,out)
 
-        def accumulateEdgeFeaturesNew(self, edgeFeatures, out=None):
+        def accumulateEdgeStatistics(self, edgeFeatures, out=None):
             if not isinstance(self, RegionAdjacencyGraph):
                 raise AttributeError("accumulateEdgeFeaturesNew not implemented for " + type(self))
             graph = self.baseGraph
             affiliatedEdges = self.affiliatedEdges
-            out = graphs._ragEdgeFeaturesNew(self, graph, affiliatedEdges, edgeFeatures, out)
+            out = graphs._ragEdgeStatistics(self, graph, affiliatedEdges, edgeFeatures, out)
             return out
 
         def accumulateNodeFeatures(self,nodeFeatures,acc='mean',out=None):
--- a/vigranumpy/src/core/export_graph_rag_visitor.hxx
+++ b/vigranumpy/src/core/export_graph_rag_visitor.hxx
@@ -162,7 +162,7 @@ public:
                     python::arg("out")=python::object()
                 )
             );
-            python::def("_ragEdgeFeaturesNew",
+            python::def("_ragEdgeStatistics",
                 registerConverters(
                     &pyRagEdgeFeaturesFromImplicit< float, float, ImplicitEdgeMap >
                 ),
@@ -689,11 +689,10 @@ public:
                 n_bins = std::max( n_bins_min, std::min(n_bins, n_bins_max) );
                 a.setHistogramOptions(HistogramOptions().setBinCount(n_bins));
                 
-                // copy values to multi array
-                MultiArray<1, double> data( Shape1(affEdges.size()) );
-                for(size_t i=0;i<affEdges.size();++i)
-                    data(i) = otfEdgeMap[affEdges[i]];
-                extractFeatures(data.begin(), data.end(), a);
+                // accumulate the values of this edge
+                for(unsigned int k=1; k <= a.passesRequired(); ++k)
+                    for(size_t i=0;i<affEdges.size();++i)
+                        a.updatePassN( otfEdgeMap[affEdges[i]], k );
                 
                 feat[0] = get<Mean>(a);
                 feat[1] = get<Sum>(a);


consti123:
I have now checked out the errors from the test servers:
The windows build failed due to a missing include of cmath.
I don't really know, why it worked on the other problems, apparently the boost header have included it already (but why not on win?).
I have added the include, let's see if this fixes the problem.
The Mac build failed due to a timeout...

consti123:
This didn't fix the problem, apparently MSVC doesn't implement is finite in cmath:
http://stackoverflow.com/questions/14580168/stdisfinite-on-msvc

ukoethe:
On MSVC, this function is called `_finite()`. To be portable, it is best to use `vigra::isinf(x)` from `mathutil.hxx`.

consti123:
Now the windows test run through.
One of the mac tests still has a timeout, the rest is working.

ukoethe:
Please merge with / rebase on branch `version-1-11-rc` to fix the test timeout problem. Thanks.

stuarteberg:
Just curious: why use `boost::accumulators` here instead of vigra's `RegionFeatures`?

ukoethe:
> Just curious: why use boost::accumulators here instead of vigra's RegionFeatures?

Because that's what Thorsten opted to use. It is, in fact, interesting to compare the accuracy of the (very different) quantile approximations of the two libraries. I hope that Constantin will be able to do this before the merge.

consti123:
I will have a look at the vigra accumulators today and do the rebase.
Will let you know about my findings about the quantile performance and the general comparison.

consti123:
I have done the rebase - let's see if the tests still work and if this fixes the timeout.

consti123:
All tests pass now :).
I am working on the comparison of the boost and vigra accumulators and will give you an update this afternoon or evening.

consti123:
Hmm due to the rebase this now contains a bunch of commits from version-1-11-rc.
Are you sure, you want to merge all of this into master or should I change the pull request to the version-1-11-rc branch?

I have now implemented everything with the vigra accumulators and made some tests to
see whether this suits my needs. In summary, it does :). (Gonna post a more extensive summary soon.)

consti123:
# Comparison of Feature Accumulation over Edges: boost and vigra accumulators

I have done some *black box* tests, to see if using vigra accumulator would be ok for 
what I need in the workflow.

## Time Performance

Calculation times for accumulation over the Edges for all datasets, I have at hand.
For all dataset I accumulate over the filters that are calculated from the raw data
with the appropriate anisotropy factor.

All calculations were done on my laptop (i7, 8 cores). 
However I havent reproduced this and did not take care of additional processes etc.
Hence these numbers are tentative, I just wanted to get a rough estimate.

| Dataset       | calc. time (boost)    | calc. time (vigra)    |
|---------------|----------------------:|----------------------:|
| test block    | 0.4 s                 | 0.4 s                 |  
| pedunculus    | 17.1 s                | 18.1 s                |
| seganiso data | 44.4 s                | 45.6 s                |
| isbi12 train  | 17.4 s                | 16.8 s                |
| isbi13 train  | 179.8 s               | 250.1 s               |

The vigra accumulator seems to be slower (probably because it needs two passes to 
compute the quantiles). However I think that this speed difference is still ok, because it is 
still a huge speedup in contrast to the old workflow.

Furthermore for large datasets the bottleneck is the calculation of the filter
we use for accumulation, not the accumulation itself.

## Random Forest performance

Performance of a Random Forest (500 trees) trained with the features calculated above.
Evaluated on a train / test - split of the data.

| Dataset       | acc-train (boost)     | acc-test (boost)      | acc-train (vigra) | acc-test (vigra)  |
|---------------|----------------------:|----------------------:|------------------:|------------------:|
| test block    | 99.93                 | -                     | 99.97             | -                 |
| pedunculus    | 99.57                 | 91.52                 | 99.55             | 91.58             |
| seganiso data | 94.26                 | 90.53                 | 94.29             | 90.67             |
| isbi12 train  | 96.39                 | 91.71                 | 96.39             | 91.96             |
| isbi13 train  | 98.45                 | 84.92                 |                   |                   |

Apparently, there is no significant difference in the performance of the Random Forest with the two different
accumulators. (Of course for a better comparison, I should have done cross validation,
but again, I was only interested in a rough estimate here).

Basically, I am fine with using the vigra accumulators, we can discuss the rest tomorrow.

jakirkham:
May want to merge/rebase on `version-1-11-rc`. It includes some more needed CI improvements.

consti123:
I have squashed my changes into a single commit now, because there was a broken commit and a spurious file (added and then removed again) in there.
Should be good to pull now (if checks run through).

stuarteberg:
Hi Constantin,

>Calculation times for accumulation over the Edges for all datasets, I have at hand.

Just curious: Roughly how large is each of those datasets you listed above?  And for your tests, were you computing the features over 3D superpixels (as you would for isotropic data), or for 2D superpixels (for anisotropic)?

If you happen to know roughly how many superpixels were in each dataset, that would be interesting, too.  (But if you can't immediately recall, don't worry about it.)

Best,
Stuart

consti123:
Hi Stuart,
for all features were calculated for 2d superpixel (because all datases liste above are anisortopic).
I can teel you the exact numbers for the smallest  (pedunculs) and largest (isbi2013) data: 
* Smallest: Volume: 512 x 512 x 30, N_superpixel ~ 5000
* Largest Volume: 1024 x 1024 x 100, N_superpixel ~ 70000

stuarteberg:
Thanks, Constantin.  One more question:

>For all dataset I accumulate over the filters that are calculated from the raw data

How many filters (channels) did you use for these tests?

consti123:
Hey Stuart,
what I have written above is not totally correct, for the largest data, I calculate the filters in 3d, but with sigma scaled in the z - direction with the anisotropy factor. (For all other datasets the calculation is purely in 2d. I do it this way, because the largest dataset has a smaller anisotropy for which this proved to be beneficial.)
This is also why the number of channels is different for both datasets (Hessian of Gaussian has 2 Eigenvalues in 2d, but 3 in 3d):
Datasets with 2d Filters: 12 channels
Largest dataset with 3d Filters: 15 channels

PR #330, opened 2016-02-22
python: Added applyMapping() function

stuarteberg:
Here's the docstring:

```
applyMapping( (object)labels, (dict)mapping [, (bool)allow_incomplete_mapping=False [, (object)out=None]]) -> object :
    Map all values in `labels` to new values using the given mapping (a dict).
    Useful for maps with large values, for which a numpy index array would need too much RAM.
    To relabel in-place, set `out=labels`.

    Parameters
    ----------
    labels: ndarray
    mapping: dict of ``{old_label : new_label}``
    allow_incomplete_mapping: If True, then any voxel values in the original data that are missing
                              from the mapping dict will be copied (and casted) into the output.
                              Otherwise, an ``IndexError`` will be raised if the map is incomplete
                              for the input data.
    out: ndarray to hold the data. If None, it will be allocated for you.
         The dtype of ``out`` is allowed to be smaller (or bigger) than the dtype of ``labels``.

    Note: As with other vigra functions, you should provide accurate axistags for optimal performance.
```

Edit: This PR uses a C++11 lambda.  If that's not acceptable (until vigra 1.11), let me know.

### DIFF
--- a/vigranumpy/src/core/segmentation.cxx
+++ b/vigranumpy/src/core/segmentation.cxx
@@ -56,6 +56,9 @@
 #include <string>
 #include <cmath>
 
+#include <boost/python/stl_iterator.hpp>
+#include <boost/unordered_map.hpp>
+
 #include "tws.hxx"
 
 namespace python = boost::python;
@@ -1188,6 +1191,72 @@ python::tuple  pyUnionFindWatershedsBlockwise(
     return python::make_tuple(out, nSeg);
 }
 
+/** \brief Map all values in src to new values using the given mapping (a dict).
+ *  See python docstring for details.
+*/
+template <unsigned int NDIM, class SrcVoxelType, class DestVoxelType>
+NumpyAnyArray
+pythonApplyMapping(NumpyArray<NDIM, Singleband<SrcVoxelType> > src,
+                   python::dict mapping,
+                   bool allow_incomplete_mapping = false,
+                   NumpyArray<NDIM, Singleband<DestVoxelType> > res = NumpyArray<NDIM, Singleband<SrcVoxelType> >())
+{
+    using namespace boost::python;
+
+    res.reshapeIfEmpty(src.taggedShape(), "applyMapping(): Output array has wrong shape.");
+
+    // Copy dict into a c++ unordered_map of ints,
+    // which is ~10x faster than using a Python dict
+    typedef boost::unordered_map<SrcVoxelType, DestVoxelType> labelmap_t;
+    labelmap_t labelmap(2*len(mapping)); // Using 2*N buckets seems to speed things up by 10%
+
+    typedef stl_input_iterator<tuple> dict_iter_t;
+
+#if PY_MAJOR_VERSION < 3
+    dict_iter_t map_iter = mapping.iteritems();
+#else
+    dict_iter_t map_iter = mapping.items();
+#endif
+
+    for (; map_iter != dict_iter_t(); ++map_iter)
+    {
+        object key = (*map_iter)[0];
+        object value = (*map_iter)[1];
+        labelmap[extract<SrcVoxelType>(key)] = extract<DestVoxelType>(value);
+    }
+
+    {
+        PyAllowThreads _pythread;
+
+        if (allow_incomplete_mapping)
+        {
+            transformMultiArray(src, res,
+                [&labelmap](SrcVoxelType px) -> DestVoxelType {
+                    typename labelmap_t::const_iterator iter = labelmap.find(px);
+                    if (iter == labelmap.end())
+                    {
+                        // Key is missing. Return the original value.
+                        return static_cast<DestVoxelType>(px);
+                    }
+                    return iter->second;
+                });
+        }
+        else
+        {
+            transformMultiArray(src, res,
+                [&labelmap](SrcVoxelType px) -> DestVoxelType {
+                    return labelmap.at(px);
+                });
+        }
+    }
+
+    return res;
+}
+
+// Unfortunately, can't use this macro because the template args uses TWO dtypes
+//VIGRA_PYTHON_MULTITYPE_FUNCTOR_NDIM(pyApplyMapping, pythonApplyMapping)
+
+
 void defineSegmentation()
 {
     using namespace python;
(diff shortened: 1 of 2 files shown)


vigranumpy/src/core/segmentation.cxx
```diff
@@ -1188,6 +1191,70 @@ python::tuple  pyUnionFindWatershedsBlockwise(
...
+#endif
+
+    for (; map_iter != dict_iter_t(); ++map_iter)
+    {
+        object key = (*map_iter)[0];
+        object value = (*map_iter)[1];
+        labelmap[extract<uint32_t>(key)] = extract<uint32_t>(value);
+    }
+
+    PyAllowThreads _pythread;
```
ukoethe: This must always go into a scope below `return res` because the latter manipulates Python objects and therefore needs the GIL (which is only acquired after the return in the present design).

stuarteberg:
I fixed the Python3 failure. (Thanks for providing CI coverage for that).  One of the XCode builds failed anyway, with the following message:

```
The job exceeded the maximum time limit for jobs, and has been terminated.
```

I'll trigger another build just for the sake of seeing green checkmarks, but I don't think that needs to hold up acceptance of the PR.

vigranumpy/src/core/segmentation.cxx
```diff
@@ -1188,6 +1191,70 @@ python::tuple  pyUnionFindWatershedsBlockwise(
...
+    dict_iter_t map_iter = mapping.iteritems();
+#else
+    dict_iter_t map_iter = mapping.items();
+#endif
+
+    for (; map_iter != dict_iter_t(); ++map_iter)
+    {
+        object key = (*map_iter)[0];
+        object value = (*map_iter)[1];
+        labelmap[extract<uint32_t>(key)] = extract<uint32_t>(value);
```
ukoethe: Shouldn't this be
```c++
labelmap[extract<SrcVoxelType>(key)] = extract<DestVoxelType>(value);
```

vigranumpy/src/core/segmentation.cxx
```diff
@@ -1188,6 +1191,70 @@ python::tuple  pyUnionFindWatershedsBlockwise(
...
+            [&labelmap](SrcVoxelType px) -> DestVoxelType {
+                return labelmap.at(px);
+            });
+    }
+
+    return res;
+}
+
+// Unfortunately, can't use this macro because the template args uses TWO dtypes
+//VIGRA_PYTHON_MULTITYPE_FUNCTOR_NDIM(pyApplyMapping, pythonApplyMapping)
```
ukoethe: Feel free to extend the macro accordingly ;-)

vigranumpy/src/core/segmentation.cxx
stuarteberg: Oh crap.  Yes!

ukoethe:
> One of the XCode builds failed

Our latest additions cause the build to exceed the 50 min timeout. I've asked Travis to extend the limit (not sure if I used the right email). Let's see how they react. A failing test means that I won't merge anything into master...

vigranumpy/src/core/segmentation.cxx
stuarteberg: Did not know that; thanks for the tip.  Fixed.

ukoethe:
> Thanks for providing CI coverage for Python3

Much of the credit should go to @jakirkham.

vigranumpy/test/test_segmentation.py
```diff
@@ -33,6 +33,49 @@ def test_labelMultiArray():
     _impl_test_labelMultiArray(numpy.uint8)
     _impl_test_labelMultiArray(numpy.uint32)
     _impl_test_labelMultiArray(numpy.float32)
+
+
+def _impl_test_applyMapping(dtype):
+    original = numpy.arange(100, dtype=dtype ).reshape(10,10)
+    mapping = { int(k) : int(k + 100) for k in original.flat[:] }
+
```
stuarteberg: BTW, Here, I'm converting from `np.uint8` (or whatever) to the standard Python `int` type to avoid errors from `boost::python::extract<>()` such as the following:

```
TypeError: No registered converter was able to produce a C++ rvalue of type unsigned char from this Python object of type numpy.uint8
```

Is there a clever way to teach `extract<>()` how to deal with `np.uint8`, etc.?  At the moment, users of `applyMapping()` will be required to convert their dict entries into `int` themselves, as I have done in this test.

stuarteberg: PS - This is occasionally a problem for other vigra functions.  For example:

```pytb
In [44]: vigra.gaussianSmoothing(a, np.float32(1.0))
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-44-3c08ae8b4e07> in <module>()
----> 1 vigra.gaussianSmoothing(a, np.float32(1.0))

TypeError: No registered converter was able to produce a C++ rvalue of type double from this Python object of type numpy.float32
```

stuarteberg:
>Much of the credit should go to @jakirkham.

:beers:

vigranumpy/test/test_segmentation.py
ukoethe: I don't know of a general solution, but in case of the labelmap, you could use the promote type:
```
typedef typename NumericTraits<SrcVoxelType>::Promote KeyType;
typedef typename NumericTraits<DestVoxelType>::Promote ValueType;
typedef boost::unordered_map<KeyType, ValueType> labelmap_t;
labelmap[extract<KeyType>(key)] = extract<ValueType>(value);
```

jakirkham:
Thanks. Though @ukoethe, @FynnBe, @bluescarni, and others did the actual Python 3 conversion, which seems like the harder part.

Yeah, so, was this the test for Mac on Python 3? That is to be expected, unfortunately.

The problem is we are using homebrew on Mac to acquire dependencies. Homebrew supplies bottles for some packages, but their bottle for Boost.Python is for Python 2 only. After some discussion with the maintainers of homebrew, they decided not to support Boost.Python for Python 3. So, on Python 3 we have to build Boost.Python from source. Unfortunately, this sometimes makes the build take a little too long and so Travis terminates it. We briefly considered a workaround where we would include a bottle of Boost.Python for Python 3, but ended up not keeping this. If you would wish to set this up, it should be too hard. If you have 10.11, you can just build the bottle and put it on Bintray for the build to retrieve.

Though the ultimate solution is to switch to conda on all platforms. This should substantially speed up all builds. One of the main reasons I have held off on this here is that we would need a new build of the Boost package so that it properly depends on python versions. It sounds like this is planned by conda, but has not occurred yet. When I tried to build Boost and VIGRA, I ended up needing to add C++11 support and make some other changes that likely won't agree with their version of Boost when it comes out. So, at present, we are building this through other means. If you are able to get there existing recipe for Boost to build without making these changes, maybe a smooth transition to conda could be completed sooner.

jakirkham:
Maybe we should just accept using someone else's bottle for the time being ( https://github.com/ukoethe/vigra/pull/332 ).

vigranumpy/test/test_segmentation.py
stuarteberg: >in case of the labelmap, you could use the promote type

That doesn't seem to work.  I don't think the problem here has to do with type widths, does it? After all, the `gaussianSmoothing()` example above is attempting to promote from float to double, which wouldn't lose any precision.

I think it has to do with the fact that `extract<>()` doesn't seem to recognize that objects of type `np.uint32` are integers.  Maybe we need to augment `converters.cxx` to explicitly support scalars of type `np.uintXX`.

ukoethe: You are right. Unfortunately, `extract` comes from boost::python, so we 
cannot easily change it.

stuarteberg: But in theory, couldn't we add a custom converter for numpy scalar types?

stuarteberg:
@ukoethe Thanks for catching the review and catching a few problems with this PR.  At this point, all tests are passing on Travis except for the above-mentioned XCode-python3 test case.

>A failing test means that I won't merge anything into master...

Hopefully a known timeout problem on Mac don't count as a "failing test"...

jakirkham:
> Hopefully a known timeout problem on Mac don't count as a "failing test"...

True. Though as we do have a fix it might be nice to get the fix in first and then retest to make sure it is in fact working in that case.

ukoethe:
> Hopefully a known timeout problem on Mac don't count as a "failing test"

Since the status badges on the main page will display "build failing" in shiny red regardless, the reason for the failure doesn't matter.

stuarteberg:
>Since the status badges on the main page will display "build failing" in shiny red regardless, the reason for the failure doesn't matter.

Oh, maybe I'm missing something. Did *this* PR cause the timeout?  My branch is based on `master`.  How is it that `master` is passing, but this PR is failing?  Is my extra code in `segmentation.cxx` just the "straw that broke the camel's back"?

jakirkham:
I think @ukoethe (correct me if I am wrong) is simply saying that we should make sure to fix the testing error in any event. If you want to try rebasing your changes on top of mine, that should put everything back into a passing state.

jakirkham:
The changes (one commit on top of `master`) can be found here ( https://github.com/ukoethe/vigra/pull/333 ).

stuarteberg:
I think I would prefer to wait until your changes are merged into master, and then I'll rebase this branch.

ukoethe:
> If you want to try rebasing your changes on top of mine, that should put everything back into a passing state.

The build of https://github.com/ukoethe/vigra/pull/333 is passing, but the switch to the boost.python bottle bought us just 5 minutes.

> Is my extra code in segmentation.cxx just the "straw that broke the camel's back"?

Exactly.

jakirkham:
> I think I would prefer to wait until your changes are merged into master, and then I'll rebase this branch.

The plan is to merge those changes into the `version-1-11-rc` branch. If that change is incorporated into `master`, it will likely be as fast-forward merge onto that commit. So, whether you rebase now or later it won't make a difference.

That all being said, I don't know if @ukoethe wants you to put your changes in that branch as well or in `master`.

vigranumpy/test/test_segmentation.py
ukoethe: We could of course implement a replacement for `extract`. In fact, the 
header `python_utility.hxx` already provides `dataFromPython()` that 
should do the right thing.

jakirkham:
> The build of #333 is passing, but the switch to the boost.python bottle bought us just 5 minutes.

Yes, both Mac builds are still quite slow. However, we weren't getting timeout errors on Mac Python 2 builds. As they are now taking about the same amount of time, this should be enough to keep us in the green for the present.

I think our long term goal to switch to conda will really be the thing that makes a noticeable difference. Using homebrew seems to just be slow regardless. I think their choice of using `tar.gz` instead of `tar.bz2` for binaries doesn't help. Though it may not be the only cause of problems.

ukoethe:
> Using homebrew seems to just be slow regardless.

Yeah, it accounts for about 15 minutes of the total time.

ukoethe:
Merged into branch `version-1-11-rc`. We can still work on the `extract` issue, but use that branch as a base.

stuarteberg:
Works for me.  Thanks! Should I close this PR, then?

vigranumpy/test/test_segmentation.py
stuarteberg: You know a lot more about `boost::python` than I do, so I might be missing something.  But I don't think we need to implement a replacement for `extract<>()`, I think we just need to register the right conversion functions with `boost::python` so that `extract<>()` works correctly for `np.uintXX` objects.

I've implemented a prototype that illustrates the concept here:
https://github.com/stuarteberg/printnum/blob/master/src/printnum.cpp

That code works, but it probably has **terrible** performance. It calls a bajillion python functions just to convert a simple integer from one type to another.  If I knew the proper incantations for doing the same thing using only the Python C-API, then that would be a faster way to implement this code.

For the record, the output of `test.py` in that repo is this:

```
$ PYTHONPATH=build python test.py
The number is: 123
The other number is: -456
```

jakirkham:
So, I looked a little closer at homebrew time spent installing and building other stuff on Mac vs Linux. It looks like build and test steps take a bit longer on Mac. The homebrew install step still takes <10min (~9min in practice). Getting it lower while still having a working build is a bit tricky, but may be possible.

vigranumpy/test/test_segmentation.py
ukoethe: You are right, this works. I've created PR 
https://github.com/ukoethe/vigra/pull/336 with an implementation. Please 
check if it solves your problem.

stuarteberg: >I've created PR #336 with an implementation. Please check if it solves your problem.

Yes, works perfectly.  Thanks, this will save a lot of little headaches.  BTW, I made a PR onto your PR.  It eliminates the now-obsolete scalar conversions from `test_segmentation.py`.

PR #332, opened 2016-02-23
[OLD] Download Boost.Python bottle from casacore's bintray

jakirkham:
Given it appears Travis's Mac Python 3 build is failing all the time, it seems better that we try to use someone's Boost.Python bottle for Python 3. Currently, the best solution I see unless we switch to conda or someone would like to set up their own Bintray is to use someone else's ( https://github.com/Homebrew/homebrew/issues/47218#issuecomment-177844653 ). They appear to have [incorporated]( https://github.com/casacore/casacore/pull/280 ) this into [`master`]( https://github.com/casacore/casacore/blob/master/.travis/install.sh#L10-L11 ) so I think we can count on this to work for the time being (until we switch to conda). I suppose this is better than simply bundling the binary in the repo ( https://github.com/ukoethe/vigra/pull/319 ) and is certainly better than just having these tests timeout.

Thoughts @ukoethe, @stuarteberg?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,7 +61,7 @@ before_script:
   - pip install nose
   - pip install numpy
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]] && [[ ${TRAVIS_PYTHON_VERSION:0:1} -eq 2 ]]; then brew install boost-python --with-python ; fi
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]] && [[ ${TRAVIS_PYTHON_VERSION:0:1} -eq 3 ]]; then brew install boost-python --without-python --with-python3 ; fi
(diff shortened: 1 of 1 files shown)


ukoethe:
I'll merge it into the `version-1-11-rc` branch when the tests run through.

jakirkham:
Ok, sounds good. Would you like me to change the PR to be against that branch instead of `master`?

ukoethe:
Yes, would be easier.

jakirkham:
Moved to this PR ( https://github.com/ukoethe/vigra/pull/333 ).

jakirkham:
Feel free to terminate the Travis builds here ( https://travis-ci.org/ukoethe/vigra/builds/111254983 ). Otherwise, this will probably take another ~1hr.

PR #333, opened 2016-02-23
Download Boost.Python bottle from casacore's bintray

jakirkham:
Replaces https://github.com/ukoethe/vigra/pull/332.

Given it appears Travis's Mac Python 3 build is failing all the time, it seems better that we try to use someone's Boost.Python bottle for Python 3. Currently, the best solution I see, unless we switch to conda or someone would like to set up their own Bintray, is to use someone else's ( https://github.com/Homebrew/homebrew/issues/47218#issuecomment-177844653 ). They appear to have [incorporated]( https://github.com/casacore/casacore/pull/280 ) this into [`master`]( https://github.com/casacore/casacore/blob/master/.travis/install.sh#L10-L11 ) so I think we can count on this to work for the time being (until we switch to conda). I suppose this is better than simply bundling the binary in the repo ( https://github.com/ukoethe/vigra/pull/319 ) and is certainly better than just having these tests timeout.

Thoughts @ukoethe, @stuarteberg?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,7 +61,7 @@ before_script:
   - pip install nose
   - pip install numpy
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]] && [[ ${TRAVIS_PYTHON_VERSION:0:1} -eq 2 ]]; then brew install boost-python --with-python ; fi
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]] && [[ ${TRAVIS_PYTHON_VERSION:0:1} -eq 3 ]]; then brew install boost-python --without-python --with-python3 ; fi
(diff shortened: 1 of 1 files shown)


jakirkham:
FWIW, I have been running the tests under my own Travis CI account. The (passing) results can be seen here ( https://travis-ci.org/jakirkham/vigra/builds/111253341 ).

jakirkham:
@ukoethe, Travis CI is passing. As this doesn't effect AppVeyor whatsoever, I think we can safely skip it.

jakirkham:
Note the build times for Mac Python 2 and 3 tests are both quite similar. Only 9s different in this case. This demonstrates that using a bottle of Boost.Python with Python 3 support makes things comparable and should avoid build timeouts.

PR #334, opened 2016-02-24
Travis Mac Build fixes

jakirkham:
Includes some build fixes for Mac that change how brew is used. In particular, this tries to avoid uninstalling any packages that could be useful. Also, it skips updating brew so that packages like Boost don't have to be downloaded and installed (one of the longer ones). Finally, cleans up a bunch build intermediates on Mac and Linux to hopefully improve memory usage. This appears to cut down the build time to ~35min for the Mac builds.


jakirkham:
Woops, got the wrong base branch. Closed in favor of this PR ( https://github.com/ukoethe/vigra/pull/335 ).

PR #339, opened 2016-02-26
python: Add new functions unique() and relabelConsecutive()

stuarteberg:
This PR adds two more functions that are helpful when dealing with huge label volumes.  This implementation of `unique()` is 3x faster than `numpy.unique()`. Similarly, this implementation of`relabelConsecutive()` is much faster than the best python implementation I could come up with.

### DIFF
--- a/vigranumpy/src/core/segmentation.cxx
+++ b/vigranumpy/src/core/segmentation.cxx
@@ -51,13 +51,14 @@
 #include <vigra/multi_convolution.hxx>
 #include <vigra/slic.hxx>
 #include <vigra/seg_to_seeds.hxx>
-
+#include <vigra/multi_pointoperators.hxx>
 
 #include <string>
 #include <cmath>
 
 #include <boost/python/stl_iterator.hpp>
 #include <boost/unordered_map.hpp>
+#include <boost/unordered_set.hpp>
 
 #include "tws.hxx"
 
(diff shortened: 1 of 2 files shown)


vigranumpy/src/core/segmentation.cxx
```diff
@@ -1147,6 +1148,65 @@ pythonApplyMapping(NumpyArray<NDIM, Singleband<SrcVoxelType> > src,
 //VIGRA_PYTHON_MULTITYPE_FUNCTOR_NDIM(pyApplyMapping, pythonApplyMapping)
 
 
+/** \brief Find unique values in the given array.
+*/
+template <class VoxelType, unsigned int NDIM>
+NumpyAnyArray
+pythonUnique(NumpyArray<NDIM, Singleband<VoxelType> > src)
+{
+    boost::unordered_set<VoxelType> labelset;
```
ukoethe: Is there a reason to prefer boost:: over std:: here?

vigranumpy/src/core/segmentation.cxx
```diff
@@ -1147,6 +1148,65 @@ pythonApplyMapping(NumpyArray<NDIM, Singleband<SrcVoxelType> > src,
...
+                         NumpyArray<NDIM, Singleband<DestVoxelType> > res = NumpyArray<NDIM, Singleband<SrcVoxelType> >())
+{
+    using namespace boost::python;
+    res.reshapeIfEmpty(src.taggedShape(), "relabelConsecutive(): Output array has wrong shape.");
+
+    {
+        PyAllowThreads _pythread;
+
+        boost::unordered_map<SrcVoxelType, DestVoxelType> labelmap;
+        auto labelmap_end = labelmap.end();
```
ukoethe: IIRC, the end iterator can be invalidated by insert. Call .end() in the lambda.

vigranumpy/src/core/segmentation.cxx
```diff
@@ -1147,6 +1148,65 @@ pythonApplyMapping(NumpyArray<NDIM, Singleband<SrcVoxelType> > src,
...
+
+VIGRA_PYTHON_MULTITYPE_FUNCTOR_NDIM(pyUnique, pythonUnique)
+
+
+/** \brief Relabel an array such that all labels are consecutive
+ * (i.e. there are no gaps in the label values used by the array)
+ * See python docstring below for details.
+*/
+template <unsigned int NDIM, class SrcVoxelType, class DestVoxelType>
+NumpyAnyArray
```
ukoethe: I think the function should also return the new max label and the mapping (converted to a Python dict).

stuarteberg:
Thanks again for the careful review.  I've updated the code to address your suggestions.

PR #340, opened 2016-03-01
Tentative bugfix for an out-of-bounds read in watershed3d.

bluescarni:
@ukoethe The sanitizer now reports clean and the test still passes.


bluescarni:
@ukoethe I implemented the changes to the ``Any`` class as discussed, the test runs cleanly with no leaks now.

PR #341, opened 2016-03-03
WIP: recursive convolution for Gaussian smoothing and derivative kernels

svenpeter42:
not for 1.11.rc but I don't want to clutter this request with other commits


svenpeter42:
the one failing test needs to be restarted. then this should be ready to be merged.

Issue #260, opened 2015-09-10
vigra redefines several symbols from boost::graph by default


stuarteberg:
`graphs.hxx` is included by some files which seem (to a user) to have nothing to do with graphs (e.g. `multi_morphology.hxx`).  That's fine, except that vigra's default behavior is to [redefine](https://github.com/ukoethe/vigra/blob/0822c132b4850162e96139b9ba81ddceb1ab8f91/include/vigra/graphs.hxx#L63) certain symbols in the `boost` namespace that prevent the user from using `boost::graph` for her own purposes, unless she knows to set `WITH_BOOST_GRAPH` before including certain vigra headers.

Can the default behavior be changed, so that vigra users don't need to know about this implementation detail?

hmeine:
How would you suggest to do that?  Looking for boost/graph.hpp being included already via the include guards?  I don't see another possibility immediately.

Interestingly the same file [already checks for boost/tuple.hpp being included](https://github.com/ukoethe/vigra/blob/0822c132b4850162e96139b9ba81ddceb1ab8f91/include/vigra/graphs.hxx#L100] in line 100, so that would not be a new pattern...

stuarteberg:
>How would you suggest to do that?

I guess I'm not sure.  I don't know this part of vigra well enough to understand why these symbols must be defined in the `boost` namespace.  Perhaps a new namespace `vigra::graphs` could be defined, and these names could be defined within it?

Something like this:

```c++
#ifdef WITH_BOOST_GRAPH
#  include <boost/tuple/tuple.hpp>
#  include <boost/graph/graph_traits.hpp>
#  include <boost/graph/properties.hpp>

namespace vigra { namespace graphs {
    // Alias the names from boost
    typedef boost::directed_tag directed_tag;
    // ... etc ...
}}

#else

namespace vigra { namespace graphs {
    // Define everything ourselves
    struct directed_tag { };
    // ... etc ...

}}

#endif

```

In all other files, we would need to replace usage of `boost::directed_tag` with `vigra::graphs::directed_tag`.

hmeine:
Another option could be to remove this header file (which is apparently an incomplete boost substitute) and just accept the dependency on boost::graph?

ukoethe:
[...]

PR #342, opened 2016-03-03
Fix for failing boost graph concept checking.

dstoe:
See also latest comment in #260.

Boost graph concept checking fails for me in test_gridgraph_BGL. This commit makes these graph functions available in the boost namespace again, such that the concept checking works.


ukoethe:
Merged into branch version-1-11-rc

PR #348, opened 2016-03-16
allow y-stride in makeBasicImageView()

hmeine:
I would like this to be reviewed, although I am quite confident myself that this is a good change.

### DIFF
--- a/include/vigra/multi_array.hxx
+++ b/include/vigra/multi_array.hxx
@@ -3423,8 +3423,8 @@ template <class T, class Stride>
 BasicImageView <T>
 makeBasicImageView (MultiArrayView <2, T, Stride> const &array)
 {
-    vigra_precondition(array.isUnstrided(),
-       "makeBasicImageView(array): array must be unstrided (i.e. array.isUnstrided() == true).");
+    vigra_precondition(array.isUnstrided(0),
+       "makeBasicImageView(array): array must be unstrided along x (i.e. array.isUnstrided(0) == true).");
     return BasicImageView <T> (array.data (), array.shape (0),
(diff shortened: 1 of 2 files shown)


ukoethe:
> I would like this to be reviewed

In the future, **all** changes should go through pull requests to avoid breaking the tests on master. (This happend yesterday due to a timeout on Travis that's hopefully fixed now). Announcing "build failing" on the front page is just too ugly.

Regarding the present change: please add a test to convince yourself and me that it works.

hmeine:
Actually, there already is a (succeeding) test, and IIRC we could both not see the problem with that.  Now I found that it actually extracts the ROI as MA, not as view.  Therefore, the roi is not actually strided.

hmeine:
OK, does the following sound like a plan?
1. I rebase this branch on current master (which changed due to my recent push) and force-push again.
2. After you merged the pull request (maybe you are waiting for the Travis CI?) into master, one of us merges master into the 1.11 branch.
Afterwards, I would only create PRs against the branch (even though I do not really see why master is not sufficient).

ukoethe:
Don't worry, I'll merge this manually on my local machine. Just have to wait for the test of my latest changes to complete.

ukoethe:
BTW, I noticed some TIFF-related branches in your fork. Is there anything we should merge into upstream?

hmeine:
Just to summarize the state again: This PR makes the existing unit test for strided makeBasicImageView()s working, i.e. non-tautologic.  Consequently, the test fails without the change to makeBasicImageView().

ukoethe:
I merged the PR into `version-1-11-rc`and about half of the tests already got through, so it's looking good.

Issue #33, opened 2011-07-19
make vigra capable of reading tiled TIFF images


hmeine:
I would like to use the vigra library on images saved from MeVisLab.

Background: MeVisLab uses a DCM/TIFF format, in which a .dcm file contains the medical meta information and a corresponding .tif file contains the image data.  However, this uses tiled TIFF format (as MeVisLab uses a tiled image representation internally to be able to handle arbitrarily large images).  This is rather esoteric in the sense that really not many programs can read tiled TIFFs.

I have had a glance at the tiled image loading documentation of libtiff, but it is quite terse.  The existing TIFFDecoderImpl in tiff.cxx is already quite complex, so I hope I won't mess things up.

What particularly struck me, was that Andrew Mihal has explicitly changed back from the strip-based, to a scanline-based loading method in 2004 (commenting that the former can be a memory hog), see the `tiff.?xx` changes in 86fda5d3b631488490961372dfcdcb4d2f8f4554.  However, I think going to a tile-based loader would revert this change.

PR #350, opened 2016-03-16
Tiled tiff loader

hmeine:
Ulli just asked about this stale branch.  (I picked up work from 2011 and made it work again, but that was already nearly a year ago.)  The big picture of this branch is described in detail(!) in issue #33.

Let's use this PR to discuss what needs to be done to include this (after releasing 1.11.0).

If I remember correctly, I was not fully happy with this yet, but one of the reasons might just have been that the z resolution (which is most likely to differ from x/y) is not present in our TIFFs at all (since they're always accompanied by a DICOM file together with all the medical information).  But that should not block inclusion of tiled TIFF reading in VIGRA.

Also, this is for the readVolume() API only, ATM.  OTOH, there are some useful cleanups and helper classes involved that could also be used for the 2D loader, which could eventually be extended as well.


ukoethe:
I change the PR status into WIP, because I'd like to discuss the bigger picture:

* The biology community has developed several incompatible TIFF variants, such as OME TIFF and ImageJ TIFF. Is there a reasonable way for VIGRA to support these variants?
* VIGRA has a dedicated API for tiled I/O. Is it useful to provide a new ChunkedArray backend for tiled TIFF?

hmeine:
WIP status makes sense.
Yes, it would be great to provide a ChunkedArray backend as well.  That is – we need to discuss how good a match these concepts really are.
However, I would consider first merging something that allows to load tiled images as a whole.  Remember (from issue #33) that while tiled and striped TIFFs both part of the same spec, and both officially supported by libtiff, they need you to use a different API, respectively, for accessing the image data.  That means that it is currently not possible to access the data, even of small images (even of images with a single tile!), with vigra (or practically any other program, for that matter).

PR #352, opened 2016-03-18
improve documentation of Random Forest code

hmeine:


include/vigra/random_forest/rf_common.hxx
```diff
@@ -490,7 +495,7 @@ class RandomForestOptions
 
     /** How many trees to create?
      *
-     * <br> Default: 255.
+     * <br> Default: 256.
```
ukoethe: If there is a mismatch between documentation and implementation here, please change the implementation: we want 255 trees because their response fits into 8 bits.

ukoethe:
included this PR into new PR https://github.com/ukoethe/vigra/pull/356

PR #355, opened 2016-03-21
HDF5File constructor: The open mode is now optional.

dagophil:
Also: Added constructor that can use string literals.

Without this commit, creating a HDF5File from a string literal
`HDF5File hfile("some_file.h5");`
will call the constructor `HDF5File::HDF5File(bool track_creation_times)` and therefore the specified file is not opened at all. Since this leads to confusion, I added a new constructor that actually opens the file (by default in ReadOnly mode).


ukoethe:
included this PR into new PR https://github.com/ukoethe/vigra/pull/356

PR #366, opened 2016-04-20
Do not compile assertions with NDEBUG flag

### LINKED ISSUE #365: Assertions are compiled and executed with NDEBUG flag
This is not the behaviour described in the documentation.

https://github.com/ukoethe/vigra/blob/master/include/vigra/error.hxx#L207-209 was probably committed by mistake.

dstoe:
Fixes #365


dstoe:
Latest commit fixes the issue as discussed: print also file names and line numbers in log messages, but do not compile assertions in the release mode.

PR #368, opened 2016-05-10
appveyor_test on current vigra clone

kkiefer:


ukoethe:
I think the appveyor tests fail because conda upgraded the hdf5 package, and the new version is incompatible with the current vigra. I'll investigate.

Issue #374, opened 2016-06-10
RandomForest.writeHDF5() seems to corrupt h5py module


stuarteberg:
The following test program consistently induces a "Bus Error" on my machine.  Strangely, the error does not occur within a vigra function, but only afterwards, when using h5py.

Using `git bisect`, I was able to identify the commit that introduced this problem: b847941d06223c9592a87505aaaa6b86239a5f43

```python
import tempfile
import numpy as np
import vigra
import h5py

tmp_dir = tempfile.mkdtemp()

# Create RF
features = np.random.random( (10, 5) ).astype( np.float32 )
labels = np.random.randint(1,3, size=(10, 1)).astype(np.uint32)    
rf = vigra.learning.RandomForest()
rf.learnRF(features, labels)

print "Writing RF..."
# If you comment out this line, the program does not crash
rf.writeHDF5(tmp_dir + '/written_by_vigra.h5', 'forest')

print "Creating new file..."
try:
    with h5py.File(tmp_dir + '/empty-file.h5', 'w') as f:
        print "Reading non-existent key. Should raise, not crash!"
        x = f['not-there']  # <---- Bus Error ??
except KeyError:
    pass

print "Never gets here..."
```

(The easiest (only?) way to reproduce the error is to attempt to access a non-existing value in an h5 file.  Normally, that is supposed to merely raise a `KeyError`, but instead the program crashes with a bus error.)

Of course, it is possible that somehow there is a bug in either `hdf5` or `h5py` that is merely exposed by this change in vigra...

This issue was previously reported in ilastik/lazyflow#211 and ilastik/ilastik#1210 by @svenpeter42.

stuarteberg:
I don't understand what, if anything, is wrong with b847941.  But indeed, reverting it seems to fix the crash. Hence, I've submitted #375.  Since the commit in question was just a performance improvement anyway (and probably not a major one), it's probably not worth investigating too much further.

stuarteberg:
Side note: This will finally allow me to update the ilastik conda recipes to use vigra 1.11, which will, in turn, allow me to merge the multicut workflow into ilastik's master branch.

ukoethe:
I can reproduce the crash using your test program, but the function `openCreateGroup_()` (which commit https://github.com/ukoethe/vigra/commit/b847941d06223c9592a87505aaaa6b86239a5f43 modifies) is not called at all. Why would reverting solve the problem?

ukoethe:
[...]

PR #375, opened 2016-06-10
Revert b847941

stuarteberg:
Revert b847941, since it seems to introduce a strange crash. See #374 for details.


ukoethe:
See https://github.com/ukoethe/vigra/pull/376 for a better fix.

PR #377, opened 2016-06-11
distinguish between old-style and new-style HDF5 error handling

ukoethe:


stuarteberg:
Looks reasonable to me.  Jeez, the HDF5 C-API seems terrible...

PR #384, opened 2016-07-20
ChunkedArrayHDF5 constructed with chunkShape from HDF5 dataset if constructed with existing HDF5 File

constantinpape:
ChunkedArrayHDF5 is now created with the chunk shape of the underlying HDF5 dataset, if it is created with the constructor for an existing file.
For this, I had to implement getChunkShape as method of HDF5File, which returns the chunk shape of the dataset.
Still missing: If the HDF5 chunk size is not a power of 2, some precondition will fail.
We could fix this by going to the next bigger power of 2 if this is the case, but I don't know where the best place for this is, yet.


constantinpape:
Now the chunk size is ceiled to a power of 2 in the constructor.
I also added a copy constructor (previously this was leading to undefined behavior).
The assignement operator is still missing (however ).
It should be implemented following the copy-swap-idiom:
http://stackoverflow.com/questions/5695548/public-friend-swap-member-function
For this we will probably need swaps for some of the members...

PR #389, opened 2016-09-26
Fix conv radius

ukoethe:


ukoethe:
Pinging @jakirkham: 

IIRC, you implemented the configuration code to test VIGRA on OSX with Python3. It seems that hombrew upgraded to boost 1.61 in the meantime, so your trick to download a valid boost-python-1.60.0 from bintray does no longer work. How can this be fixed?

jakirkham:
Probably can just upgrade to use their new [link]( https://github.com/casacore/casacore/blob/08e3ed571f2dfb8881f8f630a2dae9958ac08f53/.travis/install.sh#L17-L18 ). Though long term it is probably wise to move off of Homebrew and apt-get to conda instead.

.travis.yml
```diff
@@ -43,6 +43,7 @@ addons:
 
 before_script:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then export PATH="/usr/local/bin:${PATH}"                                                ; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update || brew update                                                           ; fi
```
jakirkham: Why was this needed?

PR #390, opened 2016-09-28
Random forest 3

ukoethe:
Contains:
* a new random forest implementation that separates algorithms from the forest data structure (new class BinaryForest) in the spirit of the LEMON graph API (should greatly simplify the creation of new RF variants)
* full parallelization of RF training and prediction
* computation of the 3-dimensional convex hull and extension of the convex hull feature to 3D


hmeine:
Could you provide a brief description of these changes, or point me to some?

ukoethe:
See top of page.

PR #401, opened 2017-01-27
fixed DLL export of HDF5ImportInfo::shape() (patch by Mario Emmenlauer)

ukoethe:


emmenlau:
I do not want to be pushy at all! But is the next vigra release already on the horizon? Then I'd plan for making the subsequent MSYS2 release. But really, its not urgent or anything! :-)

PR #402, opened 2017-01-30
Pythonbindings and improved prediction for Random Forest 3.

constantinpape:
This PR includes:
* Fixing some issues that prevented efficient parallel prediction of RF3.
* Pythonbindings for RF3
* Allowing to instantiate RF3 with any MultiArray that supports the vigra::MultiArray API.

For benchmarking of RF3, RF2 and sklearn RF see:
https://github.com/constantinpape/rf_benchmarks


constantinpape:
I have fixed the warnings from the new random forest now.
Travis and Appveyor are failing due to some broken tests that I haven't touched, I don't know what is going wrong there.

svenpeter42:
The RF3 tests fail because the API has been changed. class_vec also isn't a vector for some templates parameters afaict.

There's also something wrong with predictProbabilties. The RF loaded from the same file predicts nonsense results like [  0.00000000e+00,   1.38062969e+05,   0.00000000e+00,
           0.00000000e+00] while the old one predicts [0 1 0 0].

It's not just normalization either:
In [22]: res = proba2[:256,:256,:]/proba[:256,:256,:]

In [23]: res[np.isnan(res)] = 0

In [24]: np.unique(res)
Out[24]: 
array([      0.        ,   24200.609375  ,   24200.61132812, ...,
        136866.40625   ,  136879.40625   ,  138062.96875   ], dtype=float32)

svenpeter42:
Scratch that - it's actually normalization. For this specific case (ArgMaxVectorAcc) I've probably 'fixed' it in https://github.com/svenpeter42/vigra/tree/rf3-test.

Someone who's more familiar with the template magic used in there needs to write a proper fix though.

constantinpape:
Ah, I see , when I wrote the code I implicitly assumed that ```node_responses``` is already normalized.
Turns out it is not, so what you do seems about right.
I haven't noticed this before, because I only checked for the correct predictions, which is not affected by this.
If you make a PR to my branch I will pull and we can see if the tests run through.

svenpeter42:
The code I wrote is an ugly hack - it shouldn't be merged anywhere :-)

For the tests to run through we somehow need to use this "Acc" thing to postprocess the node response.

constantinpape:
Hmm I don't know if using this Acc template is such a good idea, after all this design was responsible for the horrible parallelisation behaviour in the first place...
Maybe we can use it instead of your code (though I think that internally it should do more or less the same).
I will check that later.

constantinpape:
Ok, I have figuered out the issue with the ```ACC``` template:
Basically this can be instantiated with two structs ```AccArgMax``` and ```AccArgMaxVector``` which will result in different value types for the ```NodeMap```.
Hence it is important to use these for accumulation (in general all this feels pretty over-engineered...).

Anyway, most test build and pass on my laptop now, except of the test for the VarImportance, which I can't compile due to some passing rvalue to reference problems.
I have commented it out for now.

constantinpape:
I have now fixed all tests, the builds on linux and windows are passing.
The mac build is currently globally suspended on travis.

constantinpape:
@ukoethe: The travis build for mac / python 3 has timed out, could you please restart it?

ukoethe:
Strange, Mac builds now take over 50 minutes. Are your tests expensive?

constantinpape:
I haven't added any new tests.
But I guess building the additional pythonbinding adds to the build time.

PR #405, opened 2017-02-17
Fix parameter of ImagePyramid::swap

jwakely:
You can't swap with a const object. This fixes a compilation failure
with GCC 7 which rejects the incorrect function even if it isn't
instantiated.


hmeine:
Apart from the fact that this indicates a missing test, this is definitely a bug fix.

PR #406, opened 2017-03-14
Lifted hierarchical clustering

constantinpape:
This PR includes:
* Hierarchical Clustering for lifted edges
* Lifting the GIL for eccentricty center functions
* Fixing HDF5 serialization for rf3


constantinpape:
I have added a commit that lfits the GIL for the dijkstra algorithms exposed to python.

constantinpape:
As usual, the tests are not behaving...
One of the mac tests is timing out (could someone with permissions please restart it?)
The linux test for python 3.2 is killed unexpectedly when running a test for ridge regression with the old random forest. I cannot reproduce this locally and I have no idea why this could be happening.
As far as I can see, this PR should not affect the code relevant to this test at all.

PR #415, opened 2017-06-15
multi_convolution: Fix for incorrect template parameter type when using unsigned int N for TinyVector SIZE. Using int M instead.

### LINKED ISSUE #414: Template parameter mismatch on multi_convolution.hxx (MultiArray's unsigned int vs TinyVector's int)
I ran into the following error when compiling Vigra with clang version 5.0.300080  (based on LLVM 5.0.300080):

```
  ../../../../include/vigra/multi_convolution.hxx:1662:9: error: no matching function for call to 'gaussianGradientMultiArray'
          gaussianGradientMultiArray(src.bindOuter(k), grad, opt);
          ^~~~~~~~~~~~~~~~~~~~~~~~~~
  ../../../../include/vigra/multi_convolution.hxx:1699:13: note: in instantiation of function template specialization 'vigra::detail::gaussianGradientMagnitudeImpl<2, unsigned char, vigra::StridedArrayTag, float, vigra::StridedArrayTag>' requested here
      detail::gaussianGradientMagnitudeImpl<N, T1>(src.expandElements(N), dest, opt);
              ^
  ../../../../include/vigra/multi_convolution.hxx:1720:5: note: in instantiation of function template specialization 'vigra::gaussianGradientMagnitude<2, unsigned char, 3, vigra::StridedArrayTag, float, vigra::StridedArrayTag>' requested here
      gaussianGradientMagnitude(src, dest, opt.stdDev(sigma));
      ^
  test.cpp:87:5: note: in instantiation of function template specialization 'vigra::gaussianGradientMagnitude<2, vigra::TinyVector<unsigned char, 3>, vigra::StridedArrayTag, float, vigra::StridedArrayTag>' requested here
      gaussianGradientMagnitude(image, gradMag, sigmaGradMag); 
      ^
  ../../../../include/vigra/multi_convolution.hxx:1569:1: note: candidate template ignored: could not match 'triple' against 'MultiArrayView'
  gaussianGradientMultiArray(triple<SrcIterator, SrcShape, SrcAccessor> const & source,
  ^
  ../../../../include/vigra/multi_convolution.hxx:1580:1: note: candidate template ignored: could not match 'triple' against 'MultiArrayView'
  gaussianGradientMultiArray(triple<SrcIterator, SrcShape, SrcAccessor> const & source,
  ^
  ../../../../include/vigra/multi_convolution.hxx:1592:1: note: candidate template ignored: could not match 'MultiArrayView' against 'MultiArray'
  gaussianGradientMultiArray(MultiArrayView<N, T1, S1> const & source,
  ^
  ../../../../include/vigra/multi_convolution.hxx:1616:1: note: candidate template ignored: could not match 'MultiArrayView' against 'MultiArray'
  gaussianGradientMultiArray(MultiArrayView<N, T1, S1> const & source,
  ^
  ../../../../include/vigra/multi_blockwise.hxx:438:58: note: candidate template ignored: could not match 'BlockwiseConvolutionOptions' against 'ConvolutionOptions'
  VIGRA_BLOCKWISE(GaussianGradientFunctor,                 gaussianGradientMultiArray,                 1, false );
                                                           ^
  ../../../../include/vigra/multi_convolution.hxx:1514:1: note: candidate function template not viable: requires at least 6 arguments, but 3 were provided
  gaussianGradientMultiArray(SrcIterator si, SrcShape const & shape, SrcAccessor src,
  ^
  ../../../../include/vigra/multi_convolution.hxx:1559:1: note: candidate function template not viable: requires at least 6 arguments, but 3 were provided
  gaussianGradientMultiArray(SrcIterator si, SrcShape const & shape, SrcAccessor src,
```

Upon further investigation, I discovered that the cause of the error is an `unsigned int` vs `int` template mismatch, when using the same dimension `unsinged int N` for both `MultiArray` and `TinyVector`. Here's an example of a section of the code where the mismatch occurs:

```
template <unsigned int N, class T1, class S1, class T2, class S2>
inline void gaussianGradientMultiArray(MultiArrayView<N, T1, S1> const & source,
                           MultiArrayView<N, TinyVector<T2, N>, S2> dest,
                           ConvolutionOptions<N> opt )
```

Notice that `unsigned int N` is being used for both `MultiArray` and `TinyVector`, and `TinyVector` should be using an `int` as `SIZE` instead: https://github.com/ukoethe/vigra/blob/master/include/vigra/tinyvector.hxx#L617

And the correct version would be: 

```
template <unsigned int N, int M, class T1, class S1, class T2, class S2>
inline void gaussianGradientMultiArray(MultiArrayView<N, T1, S1> const & source,
                           MultiArrayView<N, TinyVector<T2, M>, S2> dest,
                           ConvolutionOptions<N> opt )
```
I'm surprised that this issue has not shown up before. Also, I noticed that `multi_convolution.hxx` already contains functions where the types of the template parameters are defined correctly: https://github.com/ukoethe/vigra/blob/master/include/vigra/multi_convolution.hxx#L1692-L1697

JaimeIvanCervantes:
Updated template definitions in `multi_convolution.hxx` to fix incorrect template parameter type when using `unsigned int N` for TinyVector `SIZE`. Now using int M instead. (Fixes #414)

*NOTE:* Some functions in `multi_convolution.hxx` already have the correct template definitions, so I based my code changes from these sections, eg: https://github.com/ukoethe/vigra/blob/master/include/vigra/multi_convolution.hxx#L1692-L1697

### DIFF
--- a/include/vigra/multi_convolution.hxx
+++ b/include/vigra/multi_convolution.hxx
@@ -1426,7 +1426,7 @@ gaussianSmoothMultiArray(MultiArrayView<N, T1, S1> const & source,
                   class T2, class S2>
         void
         gaussianGradientMultiArray(MultiArrayView<N, T1, S1> const & source,
-                                   MultiArrayView<N, TinyVector<T2, N>, S2> dest,
+                                   MultiArrayView<N, TinyVector<T2, int(N)>, S2> dest,
                                    double sigma,
                                    ConvolutionOptions<N> opt = ConvolutionOptions<N>());
 
@@ -1435,7 +1435,7 @@ gaussianSmoothMultiArray(MultiArrayView<N, T1, S1> const & source,
                                   class T2, class S2>
         void
         gaussianGradientMultiArray(MultiArrayView<N, T1, S1> const & source,
-                                   MultiArrayView<N, TinyVector<T2, N>, S2> dest,
+                                   MultiArrayView<N, TinyVector<T2, int(N)>, S2> dest,
                                    ConvolutionOptions<N> opt);
 
         // likewise, but execute algorithm in parallel
@@ -1443,7 +1443,7 @@ gaussianSmoothMultiArray(MultiArrayView<N, T1, S1> const & source,
                                   class T2, class S2>
         void
         gaussianGradientMultiArray(MultiArrayView<N, T1, S1> const & source,
-                                   MultiArrayView<N, TinyVector<T2, N>, S2> dest,
+                                   MultiArrayView<N, TinyVector<T2, int(N)>, S2> dest,
                                    BlockwiseConvolutionOptions<N> opt);
     }
     \endcode
@@ -1590,7 +1590,7 @@ template <unsigned int N, class T1, class S1,
                           class T2, class S2>
 inline void
 gaussianGradientMultiArray(MultiArrayView<N, T1, S1> const & source,
-                           MultiArrayView<N, TinyVector<T2, N>, S2> dest,
+                           MultiArrayView<N, TinyVector<T2, int(N)>, S2> dest,
                            ConvolutionOptions<N> opt )
 {
     if(opt.to_point != typename MultiArrayShape<N>::type())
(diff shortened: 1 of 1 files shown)


ukoethe:
It seems that C++ compilers get more picky over time. No compiler complained about this problem so far. Thanks for fixing it. 

However, your solution of introducing a new template parameter `M` has the disadvantage that one could now pass result arrays with inappropriate `TinyVector` length. The gradient vector should always have length equal to the dimensioanlity of the original array. Thus, I'd prefer the cast `int(N)` to be used throughout (you use it at one occasion). What do you think?

JaimeIvanCervantes:
I'm now using `int(N)` instead.

I did not replace the `int M` that was already there before my commit though, should I also use `int(N)` here: https://github.com/ukoethe/vigra/blob/master/include/vigra/multi_convolution.hxx#L1692-L1697

Also, is there any reason that `TinyVector` uses `int` for `SIZE` instead of using `unsigned int` (like MultiArray)?

Issue #413, opened 2017-06-12
Test failure in vigranumpy


jakirkham:
```python
======================================================================
ERROR: Failure: ImportError (No module named 'Release')
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\bld\vigra_1497284636422\_b_env\lib\site-packages\nose\failure.py", line 39, in runTest
    raise self.exc_val.with_traceback(self.tb)
  File "C:\bld\vigra_1497284636422\_b_env\lib\site-packages\nose\loader.py", line 407, in loadTestsFromName
    module = resolve_name(addr.module)
  File "C:\bld\vigra_1497284636422\_b_env\lib\site-packages\nose\util.py", line 312, in resolve_name
    module = __import__('.'.join(parts_copy))
ImportError: No module named 'Release'
----------------------------------------------------------------------
Ran 89 tests in 8.939s
```

ref: https://ci.appveyor.com/project/jakirkham/vigra-feedstock/build/1.0.22/job/n0p01n2ydnq4xqwb#L924

jakirkham:
My best guess is this has something to do with how things work on these [lines]( https://github.com/ukoethe/vigra/blob/Version-1-11-1/vigranumpy/test/set_paths.py.in#L43-L44 ). Right now I'm creating the directory `build_Release` to perform the CMake build in. Maybe I can workaround it by using `build` as the build directory.

Still it would be nice if the former worked as it looks like it might become a pattern in conda-forge both to workaround issues causes by Google's use of the `BUILD` file for Bazel and to distinguish Release and Debug builds on Windows.

jakirkham:
It doesn't seem to have anything to do with the build directory name (have already switched to `build`). Stumped as to what could be going wrong.

ref: https://ci.appveyor.com/project/jakirkham/vigra-feedstock/build/1.0.26/job/wna6ykrh30x2gpev

ukoethe:
I've never seen an error like this. The strange thing about it is that `test_segmentation.py` only contains five tests, but the failure is reported on the sixth one, which is not supposed to exist at all. Can you find out what code `nose` is actually trying to execute here?

ukoethe:
I was able to partly reproduce the error: when I execute
```
nosetests  .  testRelease
```
I get the output
```
.....E
======================================================================
ERROR: Failure: ImportError (No module named 'testRelease')
----------------------------------------------------------------------
Traceback (most recent call last):
[...]

PR #416, opened 2017-06-24
Drop configuration from Windows Python test run

jakirkham:
Fixes https://github.com/ukoethe/vigra/issues/413

This was cause `nose` to search for a directory named after the `%CONFIGURATION%`, which simply did not exist. Hence it is dropped.

### DIFF
--- a/vigranumpy/test/CMakeLists.txt
+++ b/vigranumpy/test/CMakeLists.txt
@@ -134,7 +134,7 @@ IF(NOT PYTHON_NOSETESTS_NOT_FOUND)
         get_filename_component(PYTHON_PATH ${PYTHON_EXECUTABLE} PATH)
         get_filename_component(PYTHON_EXE  ${PYTHON_EXECUTABLE} NAME)
         SET(EXTRA_PATH  "${PYTHON_PATH}${PATHSEP}${EXTRA_PATH}")
-        SET(VIGRA_TEST_EXECUTABLE "${PYTHON_EXE} -c \"import nose; nose.main()\" . %CONFIGURATION%")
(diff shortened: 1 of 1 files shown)


jakirkham:
Appears this causes trouble for the CI build. Seems to be running into issues loading a library. However am at a loss as to why. Have tried tweaking the `PATH` and `PYTHONPATH`, but thus far have been unsuccessful at getting it to work.

ukoethe:
I don't think you can simply drop the `%CONFIGURATION%` argument in the test command line. It tells the tests whether the current build refers to a release or debug binary. Visual Studio puts these binaries into different subdirectories, and the argument is needed for the test program (specifically, line 39 of `set_path.py`) to find them. In the nose versions I'm using, non-existing tests on the command line are simply ignored (which makes sense, IMHO). Is it possible that your nose variant changed this behavior into an error? If this was the case, there might be a command line switch that recreate the old behavior.

jakirkham:
Ok, well, I think what is happening is this directory doesn't seem to exist when using `nmake`.

I don't think `nose` has changed behavior. It looks like we are using the same version of `nose` (1.3.7).

ukoethe:
> Ok, well, I think what is happening is this directory doesn't seem to exist when using nmake.

That's right. I guess, this means that your initial idea of dropping the `%CONFIGURATION%` parameter was right, only that the change needs to be restricted to `nmake` builds.

jakirkham:
Tried a different strategy in PR ( https://github.com/ukoethe/vigra/pull/417 ), which seems to fix `nmake test` without breaking the Visual Studio build. Would appreciate your feedback on it. Going to close this one out as it doesn't work for both cases.

PR #417, opened 2017-06-27
Set VIGRA_CONFIGURATION in tests always

jakirkham:
Fixes https://github.com/ukoethe/vigra/issues/413

Fixes an issue when running the tests with NMake by always setting `VIGRA_CONFIGURATION` to `CMAKE_CFG_INTDIR`.

### DIFF
--- a/vigranumpy/test/CMakeLists.txt
+++ b/vigranumpy/test/CMakeLists.txt
@@ -153,11 +153,7 @@ IF(NOT PYTHON_NOSETESTS_NOT_FOUND)
     # variable CMAKE_CFG_INTDIR contains a dot '.' for a Windows nmake build, or
     #                           '$(OutDir)' for a VisualStudio build (OutDir will be
     #                           set by VS at build time)
-    IF(NOT CMAKE_CFG_INTDIR STREQUAL ".")
-        SET(VIGRA_CONFIGURATION ${CMAKE_CFG_INTDIR})
-    ELSE()
-        SET(VIGRA_CONFIGURATION)
-    ENDIF()
+    SET(VIGRA_CONFIGURATION ${CMAKE_CFG_INTDIR})
 
     IF(AUTOEXEC_TESTS)
         add_custom_command(
--- a/config/run_test.bat.in
+++ b/config/run_test.bat.in
@@ -5,7 +5,11 @@ if [%1] NEQ [] (
     set CONFIGURATION=Release
 )
 
-echo Testing configuration '%CONFIGURATION%'
+if "%CONFIGURATION%" EQU "." (
+    echo Testing configuration '@CMAKE_BUILD_TYPE@'
+) else (
+    echo Testing configuration '%CONFIGURATION%'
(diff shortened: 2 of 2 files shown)


vigranumpy/test/CMakeLists.txt
```diff
@@ -153,7 +153,7 @@ IF(NOT PYTHON_NOSETESTS_NOT_FOUND)
     # variable CMAKE_CFG_INTDIR contains a dot '.' for a Windows nmake build, or
     #                           '$(OutDir)' for a VisualStudio build (OutDir will be
     #                           set by VS at build time)
-    IF(NOT CMAKE_CFG_INTDIR STREQUAL ".")
+    IF(NOT CMAKE_GENERATOR MATCHES "^Visual Studio")
```
ukoethe: Shouldn't the condition read
```
IF(CMAKE_GENERATOR MATCHES "^Visual Studio")
```
to reproduce the old behavior? (`IF(NOT CMAKE_CFG_INTDIR STREQUAL ".")` means "not nmake", whereas `IF(NOT CMAKE_GENERATOR MATCHES "^Visual Studio")` means "not VS").

jakirkham: Yep, you're right. It should. Yet strangely this works. This makes me wonder why dropping `%CONFIGURATION%` from the `nose` call didn't work. 😕

jakirkham: Interestingly dropping the `NOT` is causing this to fail with NMake now. 😕

ref: https://ci.appveyor.com/project/jakirkham/vigra-feedstock/build/1.0.36/job/0u7xexy35t2awbak#L2980

ukoethe:
This is indeed strange. VS-based tests sometimes work without the additional argument, if a compatible version of `vigraimpex.dll` is found somewhere in the path.

jakirkham:
Now I'm starting to wonder if we just always set `VIGRA_CONFIGURATION`, whether that will fix this.

jakirkham:
I guess the real problem is that the `run_test.bat` is setting `%CONFIGURATION%` when it was unset, which means it is adding a directory that is not there in the NMake case. Adding `.` would be an improvement as the directory exists, but it would make the [`echo` statement]( https://github.com/ukoethe/vigra/blob/Version-1-11-1/config/run_test.bat.in#L8 ) a bit odd. I have an idea of how to fix this.

ukoethe:
It might be right to drop setting a configuration by default. Another clean solution might be to pass the directory only if it exists.

jakirkham:
Ok have tried updating it. Not sure if this is more or less hacky. That said, seems to work ok for Visual Studio and NMake.

ukoethe:
I think line `echo Testing configuration Release` should read `echo Testing configuration @CMAKE_BUILD_TYPE@`, because the nmake build type is defined at configure time.

jakirkham:
Agreed. Have pushed a fix. How does that look?

PR #420, opened 2017-08-08
Fix typos in documentation

jwilk:
Found using [mwic](http://jwilk.net/software/mwic) and [anorack](https://github.com/jwilk/anorack).


ukoethe:
fixed in commit b565108e23c2859662ca071497ed18cf764cb350

PR #434, opened 2018-08-14
test/multiarray/test.cxx - added test covering shape1 issue

emmenlau:
This test identifies an error that happens when `vigra::MultiArray` is initialized with an rvalue `vigra::TinyVector<long, 1>(return type of shape)` that has been initialized with an lvalue integer.

With the error, a function `vigra::MultiArray<1u, double, std::allocator<double> > (vigra::TinyVector<long, 1>)` is generated instead of  a `vigra::MultiArray<1u, double, std::allocator<double> >` constructed with a `vigra::TinyVector<long, 1>`.

### DIFF
--- a/test/multiarray/test.cxx
+++ b/test/multiarray/test.cxx
@@ -49,6 +49,7 @@
 #include "vigra/random.hxx"
 #include "vigra/timing.hxx"
 //#include "marray.hxx"
+#include <typeinfo>
 
 using namespace vigra;
 using namespace vigra::functor;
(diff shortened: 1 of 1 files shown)


emmenlau:
Please consider this PR in order to identify and hopefully fix a potential bug in the `vigra::MultiArray` constructor with a Shape initialized with an lvalue integer.

hmeine:
I must confess that I did not understand the initial report; just from reading it, I could not make out a difference between the types, and I do not understand the phrase "With the error" (I mean – it sounds ambiguous to me – is the "error" a *bug* or an error message, is it from vigra or the compiler or added by you).

emmenlau:
Dear @hmeine, I have added issue #435 to help clarify the problem. Does that help?

ukoethe:
This is not a bug, but a C++ shortcoming. Replace
```
MultiArray<1, double> lvMA4(Shape1(width));
```
with
```
MultiArray<1, double> lvMA4(Shape1{width});
```
to fix the problem (see #435 for more details).

PR #437, opened 2018-09-04
Update skeleton.hxx

DerThorsten:
this change is needed to use the skeletonization with visual studio 2017.


ukoethe:
fixed in commit b59f93d03cea5685ee0fe9cc0065d8ce65527a64

PR #439, opened 2018-09-19
Add azure pipeline

emmenlau:
This PR adds two changes:
 - Correctly detect boost_python libraries from Conda-Forge, that use a slightly different naming scheme
 - Build vigra on Azure DevOps for Linux, Windows and MacOSX with Python 2.7 and 3.6


emmenlau:
Nice!

PR #440, opened 2018-09-19
Set up CI with Azure Pipelines

ukoethe:


kaylangan:
I'm a Program Manager on Azure Pipelines. Let me know if you have any questions or suggestions. I'm happy to help!

ukoethe:
@kaylangan Thanks for reaching out. Here is my first question: I need to compile with g++-7 on Linux. A program named `g++-7` doesn't seem to be installed in the VM. How can I configure my build to use this compiler?

kaylangan:
> @kaylangan Thanks for reaching out. Here is my first question: I need to compile with g++-7 on Linux. A program named `g++-7` doesn't seem to be installed in the VM. How can I configure my build to use this compiler?

I think you can add a script and [install using apt-get](https://gist.github.com/jlblancoc/99521194aba975286c80f93e47966dc5). That said, I'm not sure how long that'll take--if it seems to be taking too long, let me know.

ukoethe:
@kaylangan Thanks! Similar question for Windows: I'm running the VM `vs2017-win2016`, which uses VisualStudio 2017. Is there a possibility to compile with VisualStudion 2015?

kaylangan:
Would the `vs2015-win2012r2` image work, or do you require VS2015 with Windows Server 2016?
https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=vsts&tabs=yaml#use-a-microsoft-hosted-agent

ukoethe:
@kaylangan That's what I was looking for. I was just unable to find the VM's name `vs2015-win2012r2` in the documentation. Could you give me a pointer, where such things are described (or improve the docu if there isn't any :wink:)?

ukoethe:
@kaylangan Ah, I didn't realize that the center column of the table refers to the VM names. Now I have another problem: conda is not installed on the `vs2015-win2012r2` VM. How can I get it?

kaylangan:
@ukoethe can you try installing miniconda via chocolatey?

ukoethe:
@kaylangan update: choco is not in the PATH, so no progress 🤔.

ukoethe:
@kaylangan The key question may be this: Is there any possibility to debug a failing build interactively in the CI virtual machine, or in a VM with identical setup (somewhere in the cloud or by means of a local copy)? Restarting every build from scratch in a trial-and-error fashion is just really time-consuming.

kaylangan:
You'll need to install choco as admin. Unfortunately, we don't have a way to log into the machines and interactively debug.

ukoethe:
@kaylangan update: I just noticed that VS2015 (and older VS versions) are also installed in the `vs2017-win2016` VM (although the VM's name suggests otherwise), so there is actually no need to use `vs2015-win2012r2`. I'll go ahead this way.

PR #446, opened 2018-10-16
Update accumulator.hxx

vidhit:
Removal of using namespace vigra and specifying members as vigra::

### DIFF
--- a/include/vigra/accumulator.hxx
+++ b/include/vigra/accumulator.hxx
@@ -134,7 +134,7 @@ The function \ref acc::extractFeatures() "extractFeatures()" scans the data in a
     #include <vigra/multi_array.hxx>
     #include <vigra/impex.hxx>
     #include <vigra/accumulator.hxx>
-    using namespace vigra::acc;
+    vigra::acc;
     typedef double DataType;
     int size = 1000;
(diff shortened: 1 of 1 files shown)


include/vigra/accumulator.hxx
```diff
@@ -134,7 +134,7 @@ The function \ref acc::extractFeatures() "extractFeatures()" scans the data in a
     #include <vigra/multi_array.hxx>
     #include <vigra/impex.hxx>
     #include <vigra/accumulator.hxx>
-    using namespace vigra::acc;
+    vigra::acc;
```
ukoethe: This call has no effect. What are you trying to achieve?

vidhit: Ah.Ya the using namespaces which I have removed in the comments were of no use but the using namespaces which I have removed in the working code were basically done to unambiguously identify a particular member of vigra. For instance, if we write only multiarray then we may have a doubt that whether multiarray belongs to Vigra or any other library. So to explicitly specify that multiarray is a part of Vigra I had removed using namespace but yes the namespaces removed in the comments are of no use. Sorry for that I didn't realize that I even edited the comments section

emmenlau: Please close this PR. It was created by accident by a student during a learning experience.

PR #454, opened 2018-10-30
Applying clang-format 6.0.0 with .clang-format settings

emmenlau:
Please review the suggested formatting in this PR.
The relevant settings file is https://github.com/BioDataAnalysis/vigra/blob/emmenlau_clang_format/.clang-format.

To apply changes to the formatting, install clang-format >= 6.0.0 and format the source file inplace:
```
clang-format -i include/vigra/multi_array.hxx
```
To change the formatting of all sources, use something like:
```
find * \( -name '*.hxx' -o -name '*.cxx' \) -exec clang-format -i \{\} \;
```


emmenlau:
Also note that clang-format does **not** currently format comments. In my personal opinion, it would be good to also reflow and format the doxygen comments in the source code. If this is interesting, I can provide an updated formatting using a commercial tool "SlickEdit". However I do not know of any free automatic comment formatting tools, so I can only help with a one-time cleanup, and future comments would need to be manually formatted correctly.

ukoethe:
It seems that this PR got too big for github to display it properly (at least in my browser). Please pick a few representative files that allow to discuss the different aspects of your formatting policies.

From the little I could still see, I don't quite understand some of the reformatting. For example, clang-format removes the space after the `template` keyword, but adds a space after the `for` keyword. In both cases, I prefer the opposite convention, but more importantly: what's the point of replacing one (arguably arbitrary) convention by a equally arbitrary one?

I also saw some cases of a real mistake where clang-format indented function declarations more than the corresponding function body. This should not happen.

emmenlau:
I will make a separate PR with a few representative examples.

About the formatting: Most properties can be configured, of course. But most importantly I'm not hooked on the one or other format. My concern is just to find an automatic formatting tool that allows the developers to deliver a predominantly consistent code formatting.

It may be easiest if you take a look at a clang configurator, or send me more of your suggestions?

PR #461, opened 2019-01-28
CMake change: use BUILD_SHARED_LIBS standard cmake attribute

emmenlau:
Vigra currently uses a cache variable `VIGRA_STATIC_LIB` to specify static or shared library builds. In the cmake world it is more common to use `BUILD_SHARED_LIBS`. This PR replaces `VIGRA_STATIC_LIB` with `BUILD_SHARED_LIBS` and slightly simplifies the cmake build instructions.


hmeine:
From my perspective, this PR could be easily merged. It looks good and also worked for me. However, I wonder if `VIGRA_STATIC_LIB` is used by others? Should we at least document this change in a ChangeLog?

hmaarrfk:
4 years later, I'm down with merging to modernize and standardize the build process with other C/C++ packages.

However, i think i might have introduced some conflicts. Feel free to rebase (the CIs should run too!)

emmenlau:
Thanks for your work to modernize vigra. I think we can close this PR in favor of your work!

PR #469, opened 2019-07-10
changed azure pipelines

DerThorsten:
- created nightly build
- changed mac build slightly

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -31,7 +31,7 @@ jobs:
           conda create --quiet --yes --mkdir --prefix $AGENT_WORKFOLDER/envs/vigra --channel conda-forge/label/gcc7 \
                 python=$(PY_VERSION) || exit 1
           echo "Activating conda environment" || exit 1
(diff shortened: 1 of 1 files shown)


DerThorsten:
@ukoethe  it seems like vigra is curerntly failing on azure piplines on mac with a segfault

```
55: Test command: /Users/vsts/agent/2.153.2/work/1/s/build/vigranumpy/test/run_vigranumpytest.sh "."
55: Test timeout computed to be: 10000000
55: /Users/vsts/agent/2.153.2/work/1/s/build/vigranumpy/test/run_vigranumpytest.sh: line 1:  2768 Segmentation fault: 11  /Users/vsts/agent/2.153.2/work/envs/vigra/bin/python -c "import nose; nose.main()" .
55/55 Test #55: vigranumpytest ...................***Failed    0.29 sec
```
I am pretty sure If you trigger a build on master will fail with the same errors.
To avoid silent failures I would suggest to have a nightly build. This scheduled nightly build was the actual intend of this PR

hmeine:
What should be done in order to get this approved? I do not understand where the segfault comes from, but having successful macOS builds again would be nice.

PR #473, opened 2020-02-12
Fix C++ version detection for WebAssembly.

attilaolah:
The Emscripten compiler (`em++`) will refuse to output the results of the buffer unless they are flushed explicitly, or implicitly with the `std::endl`. This will result in the C++ version being detected as being the literal string:

> `stdio streams had content in them that was not flushed. you should set EXIT_RUNTIME to 1 (see the FAQ), or make sure to emit a newline when you printf etc.`

### DIFF
--- a/config/output_cplusplus_version.cxx
+++ b/config/output_cplusplus_version.cxx
@@ -6,14 +6,14 @@ int main() {
         // This is a workaround for a gcc bug:
         // http://stackoverflow.com/questions/7530047/gnu-c-macro-cplusplus-standard-conform
 #ifdef __GXX_EXPERIMENTAL_CXX0X__
-        std::cout << 201103;
+        std::cout << 201103 << std::endl;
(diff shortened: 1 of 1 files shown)


hmeine:
This change causes a problem on my macOS system:
```
CMake Error at test/multiarray/CMakeLists.txt:17 (message):
  cmake error: VIGRA_CPP_VERSION not defined yet.  Call
  VIGRA_DETECT_CPP_VERSION() from the main CMakeLists file.
```
Do you see why?

attilaolah:
Hmm, I don't see why, maybe the `endl` is actually causing problems, and we'd need to flush the buffer explicitly. I'll try that.

attilaolah:
Version detection was removed for #514 in e5cff4c42e9b086f6aefd68c2e26d84f5a95c7b0.

This is now obsolete.

PR #475, opened 2020-03-25
use tuples for numpy indexing

### LINKED ISSUE #474: Future incompatibility with numpy
In https://github.com/ukoethe/vigra/issues/436 you noticed, that in future there will be a problem with numpy compatibility:

Can this warning (issued by numpy 1.15.2) explain the error?
```
vigranumpy/vigra/arraytypes.py:1263: FutureWarning: Using a non-tuple sequence 
for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. 
In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will 
result either in an error or a different result.
```

_Originally posted by @ukoethe in https://github.com/ukoethe/vigra/issues/436#issuecomment-430752341_

Now, this does also concern parts of vigra, which will have to be adapted.
E.g. in https://github.com/ukoethe/vigra/blob/8acd73a5e0408326e2a3c1ed1c2c70ffe2b954d0/vigranumpy/lib/arraytypes.py#L1122 where `slice` is a list.

I have tried to figure out the timeline for this deprecation in numpy without success. Atm I am just a but annoyed by the warnings, which I no longer want to ignore. Are you @ukoethe still accepting contributions to vigra?

k-dominik:
Hi @ukoethe,

maybe it's the social isolation but warnings in ilastik started bothering me a lot. Some of them go back to vigra and could be fixed with the patch of this PR. While the changes here are not critical right now (I couldn't find a timeline about this deprecation) it cannot hurt to be prepared.

`numpy` currently warns with the following warning when not using tuples for
indexing:

```
vigranumpy/vigra/arraytypes.py:1263: FutureWarning: Using a non-tuple sequence
for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`.
In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will
result either in an error or a different result.
```

fixed by simply converting the index to a tuple in relevant places.

fixes #474

### DIFF
--- a/vigranumpy/lib/arraytypes.py
+++ b/vigranumpy/lib/arraytypes.py
@@ -1119,7 +1119,7 @@ class VigraArray(numpy.ndarray):
                 if slicing[k] == 0 and self.shape[k] != 1:
                     raise RuntimeError("VigraArray.withAxes(): cannot drop non-singleton axis '%s'." % self.axistags[k].key)
             permutation = AxisTags(axisinfo).permutationFromNumpyOrder()
-            res = self[slicing].transposeToNumpyOrder().transpose(permutation)
+            res = self[tuple(slicing)].transposeToNumpyOrder().transpose(permutation)
         else:
             res = self.transposeToOrder(kw.get('order'))
             if kw.get('noChannels'):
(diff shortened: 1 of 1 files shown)


hmeine:
I merged the change. The CI "failure" were just cancelled builds on macOS. Windows + Linux worked, and the code changes were minimal and looked good.

k-dominik:
thx a lot hmeine!

PR #477, opened 2020-06-22
do not install sphinx doctrees

eli-schwartz:
These files are 100% not needed at install time, and they contain unreproducible info. See https://reproducible-builds.org/ for why this matters.


hmaarrfk:
can you list out the extra files that are installed?

PR #487, opened 2021-02-08
Fix typo in installation instructions

pranavashok:
The installation instructions say that one may pass `-DDWITH_HDF5=0` in order to compile without HDF5, but this gives the following warning and does not disable compiling HDF5:

```
CMake Warning:
  Manually-specified variables were not used by the project:

    DWITH_HDF5
```

The actual option must be `-DWITH_HDF5=0`.


hmeine:
Thanks!

PR #490, opened 2021-02-08
Update osx pipelines  macOS-10.15

hmaarrfk:
maybe my builds were just halted https://github.com/ukoethe/vigra/pull/489


hmaarrfk:
it is really confusing as to why it isn't triggering a pipeline

PR #492, opened 2021-03-06
build: update cmake from 2.6 to 3.10

marco-m:
CMake 2.6 is really ancient and deprecated: newer CMake will stop
supporting it. Updating also allows to use modern CMake techniques.

This also makes it easier for other CMake projects to use VIGRA
(I stumbled on this while trying to build Hugin).

To give an idea, CMake is currently at 3.19. CMake 3.10 is what is
shipped with Ubuntu 18.04, which has been released in April 2018.

If we wanted to be more conservative, we could maybe use the CMake
version shipped with Ubuntu 16.04, but 16.04 itself will be end of life
in April 2021...

Tested on macOS 10.15 Catalina.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.6)
+cmake_minimum_required(VERSION 3.10)
 
 ##################################################
 #
--- a/config/VigraDetectThreading.cmake
+++ b/config/VigraDetectThreading.cmake
@@ -3,8 +3,6 @@
 # Source modules should use VIGRA_CONFIGURE_THREADING() from VigraConfigureThreading.cmake to set includes/preprocessor definitions.
 #
 
-cmake_minimum_required(VERSION 2.8)
-
 get_property(THREADING_IMPLEMENTATION GLOBAL PROPERTY THREADING_IMPLEMENTATION)
 if(THREADING_IMPLEMENTATION)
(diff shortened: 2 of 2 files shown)


hmeine:
Pardon my ignorance – does this change improve anything in itself? I would totally agree with you that lower versions no longer have to be supported, but as long as no never features are used, couldn't we leave the old requirements in?

marco-m:
Good question indeed :-)
I think the answer is the following mix:
- it removes the warnings when building VIGRA with a recent CMake (I see this with cmake 3.19 installed by brew on macOS):
  ```
  CMake Deprecation Warning at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 2.8.12 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
  ```
- it gives the message that the project is willing to have its cmake configuration files modernized, putting aside having to convince about bumping the version.

Having said that, one could also argue that VIGRA could bump the version the moment somebody modernizes the cmake configuration.

Accepting this PR also means that be default cmake < 3.10 will refuse to build VIGRA...

Last point: if you accept the PR, I think that it should be modified to just remove any `cmake_minimum_required()` from `config/VigraDetectThreading.cmake`, I don't think it buys anything (and it already inconsistent: 2.8 vs 2.6)...

hmeine:
Thanks for elaborating; your arguments make a lot of sense.

Yes, I think one could remove the second `cmake_minimum_required()`. Would you do so?

marco-m:
@hmeine updated

hmeine:
Thanks!

PR #497, opened 2021-08-22
add BUILD_TESTS and BUILD_DOCS CMake option variables

Crghilardi:
Hello, I was recently trying to compile `vigra` as a shared library to link with another program. I did not have `Doxygen` in my build environment but did not need documentation or tests.

This PR adds two options to the top level `CMakeLists.txt` - `BUILD_DOCS` and `BUILD_TESTS` (both ON by default) that allow a user to specify if they want to compile those directories such as:

`cmake .. -DBUILD_TESTS=OFF -DBUILD_DOCS=ON ...`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -91,6 +91,9 @@ if(MACOSX)
     set(CMAKE_MACOSX_RPATH 1)
 endif()
 
+option(BUILD_DOCS "Build documentation" ON)
+option(BUILD_TESTS "Build test programs" ON)
+
 ##################################################
 #
 #     search for dependencies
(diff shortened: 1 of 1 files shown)


hmeine:
Looks useful and clean enough to merge, even for a CMake non-expert like myself. The flag names seem to be not unusual for this purpose, based on a quick web search, so :+1:.

ArchangeGabriel:
Seems something was missed somewhere, because if I build with `-DBUILD_DOCS=OFF`, I get:
```
CMake Error at CMakeLists.txt:334 (ADD_DEPENDENCIES):
  The dependency target "doc_cpp" of target "PACKAGE_SRC_TAR" does not exist.
```

Issue #476, opened 2020-04-12
hdf5 1.12.0 support


ArchangeGabriel:
hdf5 1.12.0 introduces some API changes. This results in build error:
```
/build/vigra/src/vigra-1.11.1/src/impex/hdf5impex.cxx: In function ‘H5O_type_t vigra::HDF5_get_type(hid_t, const char*)’:
/build/vigra/src/vigra-1.11.1/src/impex/hdf5impex.cxx:193:60: error: too few arguments to function ‘herr_t H5Oget_info_by_name3(hid_t, const char*, H5O_info2_t*, unsigned int, hid_t)’                                                                                        
  193 |     H5Oget_info_by_name(loc_id, name, &infobuf, H5P_DEFAULT);
      |                                                            ^
In file included from /usr/include/H5Apublic.h:22,
                 from /usr/include/hdf5.h:23,
                 from /build/vigra/src/vigra-1.11.1/include/vigra/hdf5impex.hxx:51,
                 from /build/vigra/src/vigra-1.11.1/src/impex/hdf5impex.cxx:38:
/usr/include/H5Opublic.h:188:15: note: declared here
  188 | H5_DLL herr_t H5Oget_info_by_name3(hid_t loc_id, const char *name, H5O_info2_t *oinfo,
      |               ^~~~~~~~~~~~~~~~~~~~
make[2]: *** [src/impex/CMakeFiles/vigraimpex.dir/build.make:174: src/impex/CMakeFiles/vigraimpex.dir/hdf5impex.cxx.o] Error 1
```
There might be others, see [the migrating help](https://portal.hdfgroup.org/display/HDF5/Migrating+from+HDF5+1.10+to+HDF5+1.12#MigratingfromHDF51.10toHDF51.12-functions).

hmeine:
What do you suggest?
1. Should the configuration check that hdf5 version is <1.12?
1. Should the code be adapted to version 1.12 and the CMakeFiles require that version?
1. Should the code support older and newer versions and the CMakeFiles create a conditional `#define`?
Finally, would you be willing to supply a pull request?

ArchangeGabriel:
If I’d care only about my use case (packaging for Arch Linux), I’d say 2., but 3. might be generally better.

[...]

PR #499, opened 2021-09-28
Add compatibility for HDF5 1.12

hmaarrfk:
Closes https://github.com/ukoethe/vigra/issues/476

See builds: https://github.com/conda-forge/vigra-feedstock/pull/77

The relevant code in `H5version.h` is:

```c
#if !defined(H5Oget_info_by_name_vers) || H5Oget_info_by_name_vers == 3
  #ifndef H5Oget_info_by_name_vers
    #define H5Oget_info_by_name_vers 3
  #endif /* H5Oget_info_by_name_vers */
  #define H5Oget_info_by_name H5Oget_info_by_name3
#elif H5Oget_info_by_name_vers == 2
  #define H5Oget_info_by_name H5Oget_info_by_name2
#elif H5Oget_info_by_name_vers == 1
  #define H5Oget_info_by_name H5Oget_info_by_name1
#else /* H5Oget_info_by_name_vers */
  #error "H5Oget_info_by_name_vers set to invalid value"
#endif /* H5Oget_info_by_name_vers */
```

### DIFF
--- a/src/impex/hdf5impex.cxx
+++ b/src/impex/hdf5impex.cxx
@@ -190,7 +190,11 @@ H5O_type_t HDF5_get_type(hid_t loc_id, const char* name)
 {
     // get information about object
     H5O_info_t infobuf;
+#if H5Oget_info_by_name_vers >= 3
+    H5Oget_info_by_name(loc_id, name, &infobuf, H5O_INFO_BASIC, H5P_DEFAULT);
+#else
     H5Oget_info_by_name(loc_id, name, &infobuf, H5P_DEFAULT);
+#endif
     return infobuf.type;
 }
 


hmaarrfk:
I'll remove the WIP once the conda forge builds pass. I hope I didn't get any c-preprocessor macro syntax wrong.

hmeine:
This change looks great, because it introduces support for 1.12 while still working with older versions.

hmeine:
I would like to merge this, but I though I'd test a build without it first. Interestingly, that worked, although I have hdf5 1.12.1.  After some digging, I found that my `H5pubconf.h` contains `#define H5_USE_110_API_DEFAULT 1`.

I learned enough about hdf5 headers now, though, that I trust your patch. ;-)

hmaarrfk:
If `H4pubconf.h` is a private header, then it is likely OK to keep that define. I suspect that if it is a public header, it will give some poor developer that depends on vigra a hard time.

Is `H5pubconf.h` not inlcuded as part of this file?

hmaarrfk:
> I found that my H5pubconf.h contains #define H5_USE_110_API_DEFAULT 1.

Is this referring to a private header of yours, or to a header of Vigra's? I can't find it in the source.

PR #501, opened 2021-12-06
Add python 3.10 support

k-dominik:
Summary:

* dealt with deprecated `Iterable` import from `collections`

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -5,7 +5,7 @@
 jobs:
   - job: Linux
     pool:
-      vmImage: 'ubuntu-16.04'
+      vmImage: 'ubuntu-18.04'
     strategy:
       maxParallel: 2
       matrix:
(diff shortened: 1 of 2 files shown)


hmeine:
The main change looks good to me.  I cannot check the CI setup changes, but to me they also look like updates & cleanups that should be merged.

What about the failed checks?  CMake failed with `vigranumpy dependencies NOT found while TEST_VIGRANUMPY=1` – can that be related to these changes?

hmeine:
Do the CI pipeline changes supercede #490? Would you suggest merging that as well / before this one?

k-dominik:
Hey @hmeine,

thanks a lot for looking into this!! I'll poke around CI some more (I have avoided azure pipelines so far, so I still have to catch up there a bit). My current issue would be activating the right toolset for python 3.10 (so finding the right `vcvarsall.bat` or using the right parameters to get 14.2, and I'm not sure where to find it :D).

I'll try to get this running and will ping you again once this is worth looking at (and I formed an opinion on #490).

Cheers and have a great weekend!

hmaarrfk:
I merged in the fixes in the code from this branch  (trying to keep attribution).

I'm trying to fix the CIs in https://github.com/ukoethe/vigra/pull/514

k-dominik:
thank you @hmaarrfk,

I'll close this PR then.

PR #511, opened 2022-09-07
Minor fixes for c++17/c++20

emmenlau:


lnw:
That's certainly an improvement.  It might be slightly safer + better style to use "[] (const unsigned char x) {return std::isdigit(x);}" instead. Reference: https://en.cppreference.com/w/cpp/string/byte/isdigit

hmaarrfk:
I've also rebased https://github.com/ukoethe/vigra/pull/542 to see if things work out.

hmaarrfk:
CIs passed in #542 thank you

PR #500, opened 2021-10-16
Replace bind1st and bind2nd by bind

constantinpape:
Closes #481

PR #514, opened 2022-10-22
Fix CIs

hmaarrfk:
I spoke with ukoethe by email and he agreed that dropping python 2.7 would be ok.

Closes: https://github.com/ukoethe/vigra/issues/502

PR #522, opened 2022-10-23
Correctly assign the parameer for bind2nd

hmaarrfk:
Xref: https://github.com/ukoethe/vigra/pull/500

cc: @constantinpape, just an FYI in case you run into this stuff again

### DIFF
--- a/test/classifier/test.cxx
+++ b/test/classifier/test.cxx
@@ -533,7 +533,7 @@ struct ClassifierTest
             totalWeight = std::accumulate(a, a + data.ClassIter(ii).size(),
                                           totalWeight);
             std::transform(tmp.begin(), tmp.end(),tmp.begin(), 
-                           std::bind(std::divides<double>(), totalWeight, std::placeholders::_2));
+                           std::bind(std::divides<double>(), std::placeholders::_1, totalWeight));
             MultiArrayView<2, double> 
                 should_resp(Shp(1, data.ClassIter(ii).size()), tmp.data());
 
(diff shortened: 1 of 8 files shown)


hmaarrfk:
I want to test this in https://github.com/ukoethe/vigra/pull/514 before merging it here independently.

hmaarrfk:
yeah.... i'm pretty sure that they were all misassigned: https://cplusplus.com/reference/functional/bind2nd/

test/classifier/test.cxx
```diff
@@ -533,7 +533,7 @@ struct ClassifierTest
             totalWeight = std::accumulate(a, a + data.ClassIter(ii).size(),
                                           totalWeight);
             std::transform(tmp.begin(), tmp.end(),tmp.begin(), 
-                           std::bind(std::divides<double>(), totalWeight, std::placeholders::_2));
+                           std::bind(std::divides<double>(), std::placeholders::_1, totalWeight));
```
hmaarrfk, line 536: This is very likely the only one where the position mattered. Not sure why the other stuff didn't segfault on most platforms.

Maybe the tests were never ran?

hmeine: The old implementation with bind2nd() was battle-tested and correct, but maybe that's not what you mean.

Do you mean that I may have prematurely merged #500 without checking?  That could have been the case, sorry about that.  Usually, I reviewed the code and at least ran the test suite, but I remember that at least once, that was not possible due to technical problems on my end.  In this case, since the change was already shipped via conda-forge, I may have assumed that it was safe enough.

hmaarrfk: Right. That is what I mean. Also, i think that `_2` didn't exist and that only `_1` exists. The test suite revealed that.

hmaarrfk: Yeah.... We skip the tests at conda forge, and likely only the c library is used.

PR #523, opened 2022-10-24
Fix ci osx

hmaarrfk:


hmaarrfk:
I keep getting: 

> Application built with libpng-1.4.12 but running with 1.6.38

PR #538, opened 2023-03-07
fix osx ci :)

### LINKED ISSUE #535: OSX Pipelines broken and skipped
In https://github.com/ukoethe/vigra/pull/514/ we managed to fixup the linux and windows CIs.

We could not get the OSX ones working. 

Help reviving the OSX CI on azure would be appreciated. If you have ideas, please make PRs.

k-dominik:
okay - finally it seems to work. I'll  summarize the changes here:

* test segfaulted on OSX when linking the vigranumpy  stuff against python (see [here](https://github.com/conda-forge/vigra-feedstock/blob/34e1048c818424995997b59817bbf909f43b221b/recipe/patches/ignore_libpython_macos.diff)
* failing tests because cmake picked up some framework version of `libpng` (despite stating otherwise, this [seems to be a cmake bug](https://gitlab.kitware.com/cmake/cmake/-/issues/24348)) - resolved by changing the order of package discovery via `CMAKE_FIND_FRAMEWORK` (and `CMAKE_FIND_APPBUNDLE`).
* also the ctest wouldn't properly run the shellscripts on osx (added shebang to test scripts)

fixes #535, closes  #523, closes #490

I would also suggest to close #469 - don't think nightly builds are needed now that vigra gets some love.

My comments further down were mostly recording thoughts during debugging, and don't need to be read now anymore :)


k-dominik:
update:

When I left this, I saw some problems with `openexr` and vigra picking up different versions of `imath`, with the suspicion that `openexr` doesn't correctly specify the version of `imath`. Then I got sick. In the meantime I saw some activity... and with the latest `openexr` we are down to 3 test failures... Investigating...

k-dominik:
Hm.. okay that's strange. Locally the tests pass, but in CI, there seems to be an error related to `libpng`.

The failing tests issue

```
Application built with libpng-1.4.12 but running with 1.6.39
```

e.g. [here](https://dev.azure.com/ullrichkoethe/vigra/_build/results?buildId=425&view=logs&j=57b26b1e-8001-5296-350e-8c8e397ec380&t=4e3a7e41-3548-5232-8e42-2e037d237f5e&l=3476)

I tried finding how in the world it picks it up...

My local env is exactly the same, it also picks up the same lib, but on ci it claims version `1.4.12`

<details>

<summary>Truncated cmake output</summary>

```
# before cmake
otool -L /usr/local/miniconda/envs/vigra/lib/libpng.dylib
/usr/local/miniconda/envs/vigra/lib/libpng.dylib:
	@rpath/libpng16.16.dylib (compatibility version 56.0.0, current version 56.0.0)
	@rpath/libz.1.dylib (compatibility version 1.0.0, current version 1.2.13)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1197.1.1)

# during cmake
...
Found PNG: /usr/local/miniconda/envs/vigra/lib/libpng.dylib (found version "1.4.12") 
...
Using PNG  libraries: /usr/local/miniconda/envs/vigra/lib/libpng.dylib;/usr/local/miniconda/envs/vigra/lib/libz.dylib
...


```
</details>

hmaarrfk:
The OSX Stuff seems pretty Conda-forge specific. Maybe we should report to boost at conda-forge? or ask them for advice.

k-dominik:
how about pinging @jakirkham first - in the end that's the originator of the change.

For reference, we're discussing the following code snippet:

```cmake
    IF(APPLE)
        SET(VIGRANUMPY_LIBRARIES ${Boost_PYTHON_LIBRARY})
    ELSE()
        SET(VIGRANUMPY_LIBRARIES ${Python_LIBRARIES} ${Boost_PYTHON_LIBRARY})
    ENDIF()
```

linking agains both `Python_LIBRARIES` and `Boost_PYTHON_LIBRARY` results in a segfault. Is this a `boost.Python` bug, or something inherent to using conda to build?

hmaarrfk:
I think this is due to some static linking in Python at conda-forge. I remember having run into this somewhere else.

I'm sorry I "moved" that patch. it makes finding the original author hard. I think it was jakirkham.

config/FindVIGRANUMPY_DEPENDENCIES.cmake
```diff
@@ -149,7 +149,11 @@ IF(Python_Interpreter_FOUND)
     SET(VIGRANUMPY_INCLUDE_DIRS ${VIGRANUMPY_INCLUDE_DIRS}
         CACHE PATH "include directories needed by VIGRA Python bindings"
         FORCE)
-    SET(VIGRANUMPY_LIBRARIES ${Python_LIBRARIES} ${Boost_PYTHON_LIBRARY})
+    IF(APPLE)
```
hmaarrfk, line 152: Maybe we can add a comment pointing to this PR so that other packagers can somewhat understand why we did this.

hmaarrfk: > CONDA_TOOLCHAIN_BUILD

 I didn't know about this environment variable. If this works, lets try it.

I think only the homebrew folks might open an issue, and we can take action once they do.

Ideally, they would contribute an azure pipeline as well, but I don't know homebrew enough to build up the pipeline myself.

Issue #508, opened 2022-08-21
nose testing framework needs to be replaced


ametzler:
[nose](https://nose.readthedocs.io/en/latest/) is soon going to be in in unsupported mode (python 3.13 will remove lib2to3 which nose is using.)

The nose homepage says:

> Note to Users
>
> Nose has been in maintenance mode for the past several years and will likely cease without a new person/team to take over maintainership. New projects should consider using [Nose2](https://github.com/nose-devs/nose2), [py.test](http://pytest.org/), or just plain unittest/unittest2.

PR #545, opened 2023-09-30
Migrate to pytest

hmaarrfk:
Closes https://github.com/ukoethe/vigra/issues/508

### DIFF
--- a/ci/build_unix.sh
+++ b/ci/build_unix.sh
@@ -8,7 +8,7 @@ conda config --show
 conda create \
     --quiet --yes \
     --name vigra \
-    python=${PYTHON_VERSION} c-compiler cxx-compiler \
+    python=${PYTHON_VERSION} pytest c-compiler cxx-compiler \
     zlib jpeg libpng libtiff hdf5 fftw \
     boost boost-cpp numpy h5py nose sphinx \
     openexr lemon cmake make
--- a/ci/build_windows.bat
+++ b/ci/build_windows.bat
@@ -15,7 +15,7 @@ rem ref: https://github.com/ukoethe/vigra/issues/525
 call conda create ^
     --quiet --yes ^
     --name vigra ^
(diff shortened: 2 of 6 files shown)


eli-schwartz:
Hi, it looks like this is still installing nose in CI, and a bunch of files import functions from nose.tools. Unfortunately simply switching the test runner entrypoint is not enough to actually stop using nose, so I'm not sure how much this helps at the moment. It would be great to get the final bits and pieces cleaned up too.

hmaarrfk:
sorry you are correct. PRs welcome to address these deficiencies.

hmaarrfk:
a pr like the following:

```
diff --git a/vigranumpy/test/test_color.py b/vigranumpy/test/test_color.py
index 5551a976..fd83182b 100644
--- a/vigranumpy/test/test_color.py
+++ b/vigranumpy/test/test_color.py
@@ -38,10 +38,12 @@ import sys
 print("\nexecuting test file", __file__, file=sys.stderr)
 exec(compile(open('set_paths.py', "rb").read(), 'set_paths.py', 'exec'))
 
-from nose.tools import assert_equal, raises, assert_raises, assert_true
 import vigra
 import numpy
 
+def assert_true(x):
+    assert x
+
 _have_qt = False
 try:
     from PyQt4.QtGui import QImage
diff --git a/vigranumpy/test/test_multidef.py b/vigranumpy/test/test_multidef.py
index a993a22c..7dbe9558 100644
--- a/vigranumpy/test/test_multidef.py
+++ b/vigranumpy/test/test_multidef.py
@@ -38,7 +38,8 @@ import sys
 print("\nexecuting test file", __file__, file=sys.stderr)
 exec(compile(open('set_paths.py', "rb").read(), 'set_paths.py', 'exec'))
 
-from nose.tools import assert_equal, raises, assert_raises
+# Left over from using nose which has been deprecated
+from unittest import assertRaises as assert_raises
 import vigra
 import numpy as np
 
```

would be really appreciated.

Issue #482, opened 2020-11-17
Vigra pollutes builds with pragmas that suppress warnings


samkhal:
`config.hxx` currently uses compiler-specific warning suppression pragmas, such as [here](https://github.com/ukoethe/vigra/blob/aaaf44d240aebd0eb3539b6e105a19a984000d81/include/vigra/config.hxx#L166). This pollutes the build of anyone including this file who relies on the suppressed warnings, particularly for anyone also using `-Werror`, as this introduces compile errors that are dependent on where vigra is first included.

If suppressing these warnings from within vigra is desired, `push` and `pop` pragmas should be used to apply and reset these suppressions when entering and leaving vigra code.

hmaarrfk:
We finally got our CIs running again. if you want to make a PR to address this, I would be happy to review it.

PR #549, opened 2023-10-01
Remove pragma from config.hxx

hmaarrfk:
Closes https://github.com/ukoethe/vigra/issues/482


hmaarrfk:
This doesn't seem to be adding any significant warnings.

Issue #533, opened 2023-02-21
Forward decls in multi_handle.hxx


mkrausex:
There are two forward decls at the beginning of `multi_handle.hxx` which, according to a comment there, “Visual Studio doesn't like”, and are currently disabled via `#ifndef _MSC_VER`. However, Visual Studio 2019 in C++20 mode actually seems to _expect_ these forward decls – the code compiles only if I enable them (i.e., when I remove the `#ifndef _MSC_VER`).

What would be a good solution? I guess that for some older version of Visual Studio, these forward decls had really been a problem originally. Maybe we should only keep them disabled for all Visual Studio versions before 2019, and enable them for 2019 and later? Or should it be conditional on the C++ standard version instead?

hmaarrfk:
I'm happy to be a little more agressive with the visual studio versions to help modernize the build.

Now that we have the CIs running, we can make a few CIs with different compiler versions.

mkrausex:
Do I understand you correctly that you are in favor of enabling these forward decls unconditionally - and accepting compilation problems on old Visual Studio versions? If yes, I can prepare a PR.

hmaarrfk:
yes

PR #553, opened 2023-10-19
multi_handle.hxx: enable two forward decls also for Visual Studio

mkrausex:
(issue #533)


hmaarrfk:
Thanks, i just want to check that the builds pass with this, they seemed to before.

PR #554, opened 2023-10-19
Enforce using C++11

AdrianBunk:
Support for older standards was dropped in #507

Keeping the default CMAKE_CXX_EXTENSIONS=ON fixes #550

PR #555, opened 2023-11-05
random_shuffle removed in c++17

k-dominik:
clang seems to have actually implemented removal recently :)

example for a failed build on conda due to `random_shuffle` being removed...


hmaarrfk:
hmmm. whats our C++ support policy? do we need to guard against if define statements?

hmaarrfk:
xref: https://github.com/ukoethe/vigra/pull/554

PR #560, opened 2024-03-13
Use importlib.reload instead of imp.reload

dvzrv:
The imp module has been deprecated since Python 3.4 and is removed in Python 3.12.
Replace its use with its replacement in the importlib module.


hmaarrfk:
Thank you!

PR #563, opened 2024-08-03
remove more Nose usage

a-detiste:
This is a follow-up to https://github.com/ukoethe/vigra/pull/545 

@hmaarrfk  @eli-schwartz

### DIFF
--- a/ci/build_unix.sh
+++ b/ci/build_unix.sh
@@ -10,7 +10,7 @@ conda create \
     --name vigra \
     python=${PYTHON_VERSION} pytest c-compiler cxx-compiler \
     zlib jpeg libpng libtiff hdf5 fftw \
-    boost boost-cpp numpy h5py nose sphinx \
+    boost boost-cpp "numpy<2" h5py nose sphinx \
     openexr lemon cmake make
 
 if [[ `uname` == 'Darwin' ]];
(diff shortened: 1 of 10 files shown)


hmaarrfk:
Could you also try to address the low hanging fruit of bool8 in bumpy

I think this should help

https://github.com/numpy/numpy/issues/23755

a-detiste:
I don't really see how it's related

a-detiste:
Ok now I see...

hmaarrfk:
Remind me why this isn’t a problem on master? The nose modifications that is

hmaarrfk:
(You can set a similar upper bound to windows for numpy or if you are capable you may choose to fix the underlying issues)

a-detiste:
I don't use windows or conda or even virtualenv, I stay on the Debian testing rolling release; which is today still at Numpy 1.x. This PR looks good so far.

hmaarrfk:
I guess remains: `assert_raises`.

The remaining failures in OSX are likely due to compiler strictness I've encountered in https://github.com/conda-forge/vigra-feedstock/issues/134

a-detiste:
this needs more research ...

Is it ok to try this as you suggested here https://github.com/ukoethe/vigra/pull/545 ? (this can be an other PR)

-from nose.tools import assert_raises
+from unittest import assertRaises as assert_raises

hmaarrfk:
See #565

PR #564, opened 2024-08-04
Remove usage of nose since it really isn't in fashion anymore

hmaarrfk:
cc: @a-detiste i figured i would rip off the bandaid


hmaarrfk:
see https://github.com/ukoethe/vigra/pull/565

PR #571, opened 2024-08-10
Build with libjpeg-turbo

hmaarrfk:
we stopped supporting "jpeg" at conda-forge


hmaarrfk:
windows and linux passed.

Issue #525, opened 2022-11-07
Windows compilation error with MSVC >= 14.28


k-dominik:
I've been tracking compilation issues with VS2019 compilers (I was using `14.29`, but I've found that the problem was most likely introduced in MSVC `14.28`.

I haven't found a solution (yet) but also don't want to forget everything the next time I get to it (or someone else looks into it).

### The symptom:

following compilation error:

in the end the error occurs in MSVC STL: `algorithm(6928): error C3892: '_First': you cannot assign to a variable that is const`

more details here, note I shortened the paths a little for readability:

<details>

```
..\BuildTools\VC\Tools\MSVC\14.29.30133\include\algorithm(6928): error C3892: '_First': you cannot assign to a variable that is const
...\BuildTools\VC\Tools\MSVC\14.29.30133\include\algorithm(7050): note: see reference to function template instantiation '_BidIt std::_Insertion_sort_unchecked<_RanIt,_Pr>(const _BidIt,const _BidIt,_Pr)' being compiled
        with
        [
            _BidIt=vigra::StridedScanOrderIterator<1,vigra::UInt8,vigra::UInt8 &,vigra::UInt8 *>,
            _RanIt=vigra::StridedScanOrderIterator<1,vigra::UInt8,vigra::UInt8 &,vigra::UInt8 *>,
            _Pr=std::less<void>
        ]
...\BuildTools\VC\Tools\MSVC\14.29.30133\include\algorithm(7080): note: see reference to function template instantiation 'void std::_Sort_unchecked<_RanIt,_Fn>(_RanIt,_RanIt,__int64,_Pr)' being compiled
        with
        [
            _RanIt=vigra::StridedScanOrderIterator<1,vigra::UInt8,vigra::UInt8 &,vigra::UInt8 *>,
            _Fn=std::less<void>,
            _Pr=std::less<void>
        ]
...\BuildTools\VC\Tools\MSVC\14.29.30133\include\algorithm(7085): note: see reference to function template instantiation 'void std::sort<_RanIt,std::less<void>>(const _RanIt,const _RanIt,_Pr)' being compiled
        with
        [
            _RanIt=vigra::StridedScanOrderIterator<1,vigra::UInt8,vigra::UInt8 &,vigra::UInt8 *>,
            _Pr=std::less<void>
        ]
[...]

PR #574, opened 2024-08-11
Fix the implementation of unique

hmaarrfk:
Fixes: https://github.com/ukoethe/vigra/issues/525


hmaarrfk:
Alright a review is requested here.

Not the greatest since this uses a temporary buffer when sort is requested, but it seems to be fine..

k-dominik (approved):
No matter how inelegant you might feel about the additional copy, you are my hero for fixing this :)

PR #582, opened 2024-09-02
Find sphinx better

hmaarrfk:
Sometimes in cross compiled environments sphinx may not be under the python prefix.

It should just be availbe in the bin directory


hmaarrfk:
Sphinx will try to import code for the generation so it might need a consistent environment.

PR #583, opened 2024-09-02
Update Version 1.12.1

hmaarrfk:
```
    // Convenience Macros for stringifying the version number
    #define __VIGRA_STRINGIFY(s) #s
    #define _VIGRA_STRINGIFY(s) __VIGRA_STRINGIFY(s)
    #define VIGRA_VERSION_STRING (_VIGRA_STRINGIFY(VIGRA_VERSION_MAJOR) "." _VIGRA_STRINGIFY(VIGRA_VERSION_MINOR) "." _VIGRA_STRINGIFY(VIGRA_VERSION_PATCH)

```

We should consider a stringification macro, but I don't want to wait that long.


hmaarrfk:
Given the patch was so "small" reasonable is 1 day.

PR #584, opened 2024-09-03
Removing executable permission from plain files

emmenlau:
This PR has only a trivial change: It removes executable permissions from plain files (like images or C++ source files).


k-dominik (approved):
lgtm

PR #585, opened 2024-09-03
Remove superfluous whitespace at end of line

emmenlau:
This PR contains a pure whitespace change. It removes all trailing whitespace from c++ source files. This improves future PRs, to reduce modification of whitespace.


k-dominik:
Hi @emmenlau,

also thank you for this change proposal.

While I like it tidy, personally, I also like my in-line git history.  I know there are hoops to hide certain commits from git blame but I prefer not to set this up. So far superfluous whitespace has not been a concern when working with this library. So I'd be in favor of _not_ adding this kind of bulk change now. But let's see what @hmaarrfk thinks.

Cheers
Dominik

hmaarrfk:
Let's jump through the hoops for this. 

It's impossible to do any work on this project on modern editors that keep stripping whitespace. 

I wonder if it is worth adding something like clang tidy. But that might be overkill. 

If you are ok making a followup PR with the git rev ignore I'll mergr this as is

emmenlau:
Actually, I'm all for adding something like `.clang-format` and/or clang-tidy!! We use that on all projects nowadays, and its even supported out of the box in Visual Studio and the likes!

But I can say that I tried `.clang-format` a long time ago in a MR, which eventually was never merged because the changes where quite big. I guess it would help to fiddle a bit with the config, to make sure the `.clang-format` respects the current vigra formatting as much as possible, or at least represents your idea of nice formatting :-)
This is an aspect where I'm not too knowledgeable.

Can you say that you mean with `git rev ignore`? I'd be more than happy to push this forward, and even more if we can add something like `.clang-format`!

hmaarrfk:
can you configure `.clang-tidy` to `just check trailing whitespace`? Nothing else for now.

We can reformat things slowly as you get a few more PRs through.

I want to strike the right balance between making it easy for me to review (i dislike mentioning to others to strip whitespace) and destroying the "knowledge" of the historic code base.

emmenlau:
> can you configure `.clang-tidy` to `just check trailing whitespace`? Nothing else for now.

I would have assumed that such a check should exist in `clang-tidy`. But after 5 minutes of googling I find not a single useful resource. So I conclude that `clang-tidy` does not feature checks for trailing whitespace (or consistent whitespace in general).

It should be easy to craft some shell script with such checks, in case you want to run them as part of the CI. I could craft the script, but I'm not too knowledgeable with CI integration.

hmaarrfk:
Can you:

1. Create a file called `run_lint.sh`
2. Put the following content in:
```bash
#!/bin/bash

# Find all .c, .h, and .cpp files in the current directory
files=$(find . -name '*.c' -o -name '*.h' -o -name '*.cpp')

# Initialize a flag to track if any trailing whitespace is found
error_found=0

# Loop through each file and check for trailing whitespace
for file in $files; do
    if grep -q "[[:blank:]]$" "$file"; then
        echo "Trailing whitespace found in: $file"
        error_found=1
    fi
done

# If trailing whitespace was found, exit with an error
if [ $error_found -eq 1 ]; then
    exit 1
else
    echo "No trailing whitespace found."
fi
```
3. Make the following mods to the azure pipeline

```patch
diff --git a/azure-pipelines.yml b/azure-pipelines.yml
index d8db9e68..3e13928e 100644
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -13,6 +13,11 @@ jobs:
           PYTHON_VERSION: 3.12
     steps:
       - script: bash ci/build_unix.sh
+  - job: lint
+    pool:
+      vmImage: ubuntu-latest
+    steps:
+      - script: bash run_lint.sh
   - job: OSX
     pool:
       vmImage: macOS-latest
```

4. Push
5. Let the pipeline fail
6. Fix the remaining whitespace with a script similar to:

```
$ cat strip.sh
#!/bin/bash

# Check if a filename was provided as an argument
if [ -z "$1" ]; then
    echo "Usage: $0 <filename>"
    exit 1
fi

# Check if the provided file exists
if [ ! -f "$1" ]; then
    echo "Error: File '$1' not found."
    exit 1
fi

# Remove trailing whitespace in place
sed -i 's/[[:blank:]]\+$//' "$1"

echo "Trailing whitespace removed from: $1"
```


Thanks!

hmaarrfk:
(I tried to push to your branch, but it didn't seem like you enabled maintainers to push)

emmenlau:
I seem unable to give you access to this PR! I looked at https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork but it does not help.

hmaarrfk:
yeah its fine

It seems like the CI test is working. 

Just a few more files with trailing whitespace.

emmenlau:
Is the test error on osx unrelated?

hmaarrfk:
Lets see what happens on the main branch, i just merged https://github.com/ukoethe/vigra/pull/586

hmaarrfk:
why? i was just restarting the builds.

hmaarrfk:
it seems like they passed. some tests are likely just flaky, and always have been:
image

emmenlau:
> why? i was just restarting the builds.

Sorry, I mis-read! When you said you merged, I assumed you merged this PR! :-)

emmenlau:
> can you make a PR to add the git revision to some git ignore blame or something?>

Hmm, I'm not sure I understand! Can you elaborate?

hmaarrfk:
https://graphite.dev/guides/git-blame-ignore-revs#step-2-creating-a-git-blame-ignore-revs-file

emmenlau:
Ok, I just learned about ignoring git commits, which is interesting! But, can you elaborate which commit(s) you would want to ignore, and why? I'm not sure I get the bigger picture...

hmaarrfk:
We want the commit that changed 20000 lines of code to be ignored so we can maintain the meaningful history of thibgs

PR #586, opened 2024-09-03
Ensure the header 'windows.h' is referencing 'vigra/windows.h'

emmenlau:


k-dominik:
Hi @emmenlau,

thank you so much for adding this fix. Just for my understanding, this has been a problem when building against vigra and your own project also contains a `windows.h`?

PR #589, opened 2024-09-17
Add whitespace cleanup to git blame

hmaarrfk:


hmaarrfk:
Before 
image

hmaarrfk:
After:
image

k-dominik:
normal git seems not to be as smart about merge commits as your git blame thing (although hard to tell, your second screenshot cuts off before the interesting bit) -> #590

Issue #529, opened 2023-01-20
build-error against python 3.11 - re.error: global flags not at the start of the expression


ametzler:
Good morning,

building against python 3.11 fails with:
~~~
Postprocessing html files
cd /dev/shm/VIGRA/libvigraimpex-1.11.1/obj.x86_64-linux-gnu/docsrc && /usr/bin/python3 /dev/shm/VIGRA/libvigraimpex-1.11.1/docsrc/makeFunctionIndex.py /dev/shm/VIGRA/libvigraimpex-1.11.1/doc/vigra
cd /dev/shm/VIGRA/libvigraimpex-1.11.1/obj.x86_64-linux-gnu/docsrc && /usr/bin/python3 /dev/shm/VIGRA/libvigraimpex-1.11.1/docsrc/post.py /dev/shm/VIGRA/libvigraimpex-1.11.1/doc/vigra 1.11.1
Traceback (most recent call last):
  File "/dev/shm/VIGRA/libvigraimpex-1.11.1/docsrc/post.py", line 84, in <module>
    headingSummary = re.compile(r'''(<!-- Generated by Doxygen .+ -->
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/re/__init__.py", line 227, in compile
    return _compile(pattern, flags)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/re/__init__.py", line 294, in _compile
    p = _compiler.compile(pattern, flags)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/re/_compiler.py", line 743, in compile
    p = _parser.parse(p, flags)
        ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/re/_parser.py", line 980, in parse
    p = _parse_sub(source, state, flags & SRE_FLAG_VERBOSE, 0)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/re/_parser.py", line 455, in _parse_sub
    itemsappend(_parse(source, state, verbose, nested + 1,
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #594, opened 2024-12-04
Move away from inline global regex flag

parona-source:
Fixes: https://github.com/ukoethe/vigra/issues/529


hmaarrfk:
can you merge from main? the CIs should be fixed. i couldn't push to your branch.

hmaarrfk:
thanks!

PR #595, opened 2024-12-06
CI: add python 3.13

t0b3:
this adds python 3.13 on linux
nb: macos fails for different reasons


k-dominik (approved):
Looks good, also nice to see that it works :)

hmaarrfk:
3.13 CI passed.

Issue #596, opened 2025-01-23
Build failure gcc-15.0.1 no member named ‘operator()’


brunopostle:
I'm seeing this error building on fedora rawhide ([vigra-build.log](https://github.com/user-attachments/files/18526253/vigra-build.log)):

gcc-15.0.1
boost-1.83.0
numpy-2.2.1
python-3.13.1

```
In file included from /builddir/build/BUILD/vigra-1.11.2-build/vigra-1.11.2/include/vigra/adjacency_list_graph.hxx:49,
                 from /builddir/build/BUILD/vigra-1.11.2-build/vigra-1.11.2/include/vigra/graph_algorithms.hxx:56,
                 from /builddir/build/BUILD/vigra-1.11.2-build/vigra-1.11.2/include/vigra/skeleton.hxx:45,
                 from /builddir/build/BUILD/vigra-1.11.2-build/vigra-1.11.2/vigranumpy/src/core/accumulator-region-singleband.cxx:40:
/builddir/build/BUILD/vigra-1.11.2-build/vigra-1.11.2/include/vigra/random_access_set.hxx: In member function ‘vigra::RandomAccessSet<Key, Compare, Alloc>::const_iterator vigra::RandomAccessSet<Key, Compare, Alloc>::insert(const_iterator, const value_type&)’:
/builddir/build/BUILD/vigra-1.11.2-build/vigra-1.11.2/include/vigra/random_access_set.hxx:403:46: error: ‘class vigra::RandomAccessSet<Key, Compare, Alloc>’ has no member named ‘operator()’; did you mean ‘operator=’? [-Wtemplate-body]
  403 |    if((position == begin() || this->operator()(*(position-1),value))
      |                                              ^
      |                                              operator=
/builddir/build/BUILD/vigra-1.11.2-build/vigra-1.11.2/include/vigra/random_access_set.hxx:404:44: error: ‘class vigra::RandomAccessSet<Key, Compare, Alloc>’ has no member named ‘operator()’; did you mean ‘operator=’? [-Wtemplate-body]
  404 |    && (position == end() || this->operator()(value, *position))) {
      |                                            ^
      |                                            operator=
```

brunopostle:
Workaround for gcc-15 seems to be to build with `-Wno-template-body`

k-dominik:
Fix on the way in #599

PR #599, opened 2025-02-27
fix RandomAccessSet::insert(pos, val) function for gcc-15 and clang-19

lnw:
This function has never worked, but newer compilers (gcc-15, clang-19) emit an error even while not fully instantiating this specific function, while previous versions did not.  This PR aims to correctly implement the function in question.

Fixes issue #596 .


k-dominik (approved):
Awesome, was looking into this right now and cool to find it fixed already! CI error seems unrelated!

PR #605, opened 2025-03-15
modernise for c++20

lnw:
Fix all compile time errors and warnings for clang++-19  / c++-20, while also compiling as c++-11/14/17 without errors/warnings.  g++-14 compiles as c++-11/14/17/20 without errors but several warnings (that existed previously). 

The fixed errors are mainly removed member types of std::allocator.


hmaarrfk:
Thanks!

PR #606, opened 2025-04-23
CMakeLists: fix finding boost_python

barracuda156:
Otherwise `boost_python` is not being found


hmaarrfk:
hmm is this true? and still needed? if you could show a diff of the build logs that would be great so i can better understand.

sorry i let this dorman for 1 year....

PR #607, opened 2025-04-23
random.hxx: fix macOS version macros

barracuda156:
Current macros have a wrong effect, since neither `__MAC_OS_X_VERSION_MAX_ALLOWED` nor `MAC_OS_X_VERSION_10_12` are defined on earlier systems. `AvailabilityMacros` uses macros without leading underscores. `MAC_OS_X_VERSION_10_12` is defined on 10.12+.
Fix them, so that it actually works as intended.


hmaarrfk:
um ok. but this is really for old systems. i hope you don't have to support those for too long.

barracuda156:
Thank you for merging!

Issue #609, opened 2025-07-17
Add `unit` property to `AxisInfo`


btbest:
Hi all, we're working on handling physical pixel dimensions in ilastik. The natural place to keep this metadata would be the `AxisTags`, which currently have a good place for the pixel size in `AxisInfo.resolution`, but there's no good place for the unit (the "micrometer" string) to go. We could abuse the `AxisInfo.description` for it, given that it's just an arbitrary unused string, but it would be better to have an appropriate name for the property.

We'd like to see if people would be open to extending the AxisInfo with another string property `AxisInfo.unit`. We wouldn't need any logic around the property. Probably better if there is none, as we want to be able to just pipe through whatever string we read out of input files and carry them over unchanged when exporting files.

We tried subclassing `AxisInfo` and simply calling the AxisTags constructor with `AxisTags(List[UnitAxisInfo])`. This doesn't solve the issue, because it looks like the constructor doesn't just keep references to the objects passed into it, and instead constructs new `AxisInfo` objects with only the known properties. We'd have to essentially rewrite most of the AxisTags logic in a `UnitAxisTags` subclass - pointless. We'd rather just carry around a simple axiskey-unitstring map next to the AxisTags in that case.

btbest:
We've [abandoned](https://github.com/ukoethe/vigra/pull/610#issuecomment-3162663426) the idea because it turns out the `AxisTags` constructor isn't forward-compatible - it errors if there are unknown properties in JSON-serialized AxisTags. This means that adding the `unit` property would turn all future files that store serialized AxisTags into a new format that isn't compatible with older versions of vigra.

A better first step would be to change the constructor to be permissive to additional unexpected arguments, so that at an undetermined point in the future one can extend the format.

PR #610, opened 2025-08-05
Add AxisInfo units

Peter-Laszcz:
Adds a `unit` string field and relevant methods to `AxisTags`' `AxisInfo`, in order to store unit scales corresponding to the `resolution` field. This would be a direct resolution to #609 , as well as allowing physical sizes to be implemented more effectively throughout Vigra use cases.

### DIFF
--- a/include/vigra/axistags.hxx
+++ b/include/vigra/axistags.hxx
@@ -315,7 +315,7 @@ class AxisInfo
 
     static AxisInfo c(std::string const & description = "")
     {
-        return AxisInfo("c", Channels, 0.0, description);
+        return AxisInfo("c", Channels, 0.0, description, "");
     }
 
     std::string key_, description_, unit_;


vigranumpy/src/core/axistags.cxx
```diff
@@ -564,6 +564,8 @@ void defineAxisTags()
          "    :type typeFlags: :class:`~vigra.AxisType`\n"
          "    :param resolution: the resolution (step size) of the axis\n"
          "                       (e.g. 0.0 means 'unknown')\n"
+         "    :param unit: the unit of measurement corresponding to resolution of the axis\n"
+         "                       (can be left undefined)\n"
```
k-dominik, line 568: the docstring needs more updating. 

1. in line 558, the docstring does not include the unit (it should!)
1. The position of the unit arg should be after the description (logically it belongs to resolution, but the order of arguments matters. `type = vigra.AxisInfo("x", vigra.AxisType.Space, 0, "description")` still works as expected. Reading the docstring I'd have expected to set `unit` to the text `description`.
2. Further down, there are a lot of descriptions of the factories, these need updating with unit, too.

k-dominik (commented):
Hello @Peter-Laszcz,

great to see this coming to fruition.

The CI failure seems unrelated - investigating in #611 ...

I think it's implemented soundly, following mostly how description is implemented, too.

I think the documentation could be improved, see review comments.

A maybe finer point is whether the operators, and functions like `is_compatible` should take the unit string into account. Currently, `==` comparison would equal `True` even if units mismatch. It's probably safest to keep this behavior, but it should be documented.

vigranumpy/test/test_arraytypes.py
```diff
@@ -457,35 +457,39 @@ def checkCompatibility(obj, compatible):
 
 def testAxisTags():
     axistags = AxisTags(AxisInfo.c(description="RGB"),
-                        AxisInfo.ft(3.0, "time frequency"),
-                        AxisInfo.y(0.5),
-                        AxisInfo.z(4, "confocal depth"))
+                        AxisInfo.ft(resolution=3.0, description="time frequency", unit="sec"),
```
btbest, line 460: If this is an ft axis, wouldn't an appropriate example unit be `Hz`? :)

vigranumpy/src/core/axistags.cxx
```diff
@@ -646,7 +650,10 @@ void defineAxisTags()
              "contains the given axistype.\n")
         .def("compatible", &AxisInfo::compatible,
              "\naxisinfo1.compatible(axisinfo2) yields True when the two axisinfo objects "
-             "have the same keys and types, or either of the two is 'unknown'.\n")
+             "have the same keys and types, or either of the two is 'unknown'.\n"
+             "Note that, as this does not compare axis resolutions or units, \n"
+             "there may still be mismatch or incompatibility in that regard.")
+        //Equality comparisons (note does not compare axis resolutions or units)
         .def(self == self)
```
k-dominik, line 657:
```suggestion
        .def("__eq__", &AxisInfo::operator==, "Equality comparison. Note: does not compare axis resolutions or units.")
```
Doing this for equality is probably good enough, given that there wasn't any mention before.

include/vigra/axistags.hxx
```diff
@@ -247,62 +263,62 @@ class AxisInfo
...
-    static AxisInfo ft(double resolution = 0.0, std::string const & description = "")
+    static AxisInfo ft(double resolution = 0.0, std::string const & description = "", std::string const & unit = "")
     {
-        return AxisInfo("t", AxisType(Time | Frequency), resolution, description);
+        return AxisInfo("t", AxisType(Time | Frequency), resolution, description, unit);
     }
 
     static AxisInfo c(std::string const & description = "")
     {
         return AxisInfo("c", Channels, 0.0, description);
```
btbest, line 318: This line instantiates channel axes with 0.0 resolution - shouldn't it also do `""` for the unit then?

vigranumpy/src/core/axistags.cxx
```diff
@@ -646,7 +650,10 @@ void defineAxisTags()
              "contains the given axistype.\n")
         .def("compatible", &AxisInfo::compatible,
              "\naxisinfo1.compatible(axisinfo2) yields True when the two axisinfo objects "
```
k-dominik, line 652:
```suggestion
             "\naxisinfo1.compatible(axisinfo2) yields True when the two axisinfo objects\n"
```

just because the current rendering is a bit ugly:

```
 |  compatible(...)
 |      axisinfo1.compatible(axisinfo2) yields True when the two axisinfo objects have the same keys and types, or either of the two is 'unknown'.
 |      Note that, as this does not compare axis resolutions or units,
 |      there may still be mismatch or incompatibility in that regard.
```

